PR #48, opened 2025-07-20
Implement suit puncturing

### LINKED ISSUE #4: Feature: Suit Puncturing
In order to make space a bit more dangerous and discourage long-term suit usage, damage to suits should physically affect their ability to stay pressurized.

Requirements:
- (OPTIONAL) We may want suits to actually have some level of "pressurization" simulation. Very optional for the time being though. More of a pipe-dream.
- Should have several different levels of damage
  - Minor pressure leakage (low damage)
  - Major pressure leakage (oh shit damage)
  - Complete destruction (no protection, temp leakage too)
- Damage levels should have examine text, (potentially) sprites, and popups on change.
- Repair items for patching damage
  - Items should have a max damage threshold. For example, major damage is repairable only with a sewing kit whereas minor damage can be patched with duct tape.

Partmedia:
## About the PR
Implement basic suit puncturing.

## Why / Balance
 Closes #4 .

## Technical details
I suggest rebase-merge instead of squash merge. Changes are summarized in commit history:

- Add VV fields, not related to this PR, but I used this while debugging it
- Pass positive damage to worn clothing see TODO in comments
- Implement puncture damage in pressure-protecting suits

## Media
NT reminds employees to check integrity of PPE before walking in space:

https://github.com/user-attachments/assets/feae36eb-a4d7-407c-85fe-95d903ff5e61

### DIFF
--- a/Content.Server/Atmos/EntitySystems/BarotraumaSystem.cs
+++ b/Content.Server/Atmos/EntitySystems/BarotraumaSystem.cs
@@ -5,6 +5,7 @@ using Content.Shared.Alert;
 using Content.Shared.Atmos;
 using Content.Shared.Damage;
 using Content.Shared.Database;
+using Content.Shared.Examine;
 using Content.Shared.FixedPoint;
 using Content.Shared.Inventory;
 using Content.Shared.Inventory.Events;
@@ -29,6 +30,8 @@ namespace Content.Server.Atmos.EntitySystems
             SubscribeLocalEvent<PressureProtectionComponent, GotUnequippedEvent>(OnPressureProtectionUnequipped);
             SubscribeLocalEvent<PressureProtectionComponent, ComponentInit>(OnUpdateResistance);
             SubscribeLocalEvent<PressureProtectionComponent, ComponentRemove>(OnUpdateResistance);
+            SubscribeLocalEvent<BarotraumaComponent, DamageChangedEvent>(OnDamaged);
+            SubscribeLocalEvent<PressureProtectionComponent, ExaminedEvent>(OnExamine);
 
             SubscribeLocalEvent<PressureImmunityComponent, ComponentInit>(OnPressureImmuneInit);
             SubscribeLocalEvent<PressureImmunityComponent, ComponentRemove>(OnPressureImmuneRemove);
@@ -194,6 +197,13 @@ namespace Content.Server.Atmos.EntitySystems
             highModifier = ev.HighPressureModifier;
             lowMultiplier = ev.LowPressureMultiplier;
             lowModifier = ev.LowPressureModifier;
+
+            // If damaged beyond compromised threshold, we don't offer any protection.
+            if (TryComp<DamageableComponent>(ent.Owner, out var damage)) {
+                if (damage.TotalDamage > comp.CompromisedDamageThreshold) {
+                    return false;
+                }
+            }
             return true;
         }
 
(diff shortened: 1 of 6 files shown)


Content.Server/Atmos/EntitySystems/BarotraumaSystem.cs
```diff
@@ -287,5 +297,24 @@ public override void Update(float frameTime)
...
+
+        private void OnDamaged(EntityUid uid, BarotraumaComponent comp, DamageChangedEvent args)
+        {
+            // We actually only care that the entity with an equipped PressureProtectionComponent has been damaged. But
+            // since there's no (easy?) way to get the wearer of an item, and since we know that if the worn item is
+            // damaged then the wearer must be too, have the wearer subscribe to this event instead.
+            UpdateCachedResistances(uid, comp);
+        }
+
+        private void OnExamine(Entity<PressureProtectionComponent> ent, ref ExaminedEvent args)
```
EmoGarbage404, line 315: This event subscription should be moved to a shared system otherwise the examine box will mispredict. The data here isn't reliant on any atmos data so you should just make an `ESSharedBarotraumaSystem` in the `_ES` directory and put this method in there.

Content.Shared/Damage/Systems/DamageableSystem.cs
```diff
@@ -191,6 +192,17 @@ public void DamageChanged(EntityUid uid, DamageableComponent component, DamageSp
             if (before.Cancelled)
                 return null;
 
+            // Hit anything the target is wearing, too. But on recursive hits, only pass positive values. Healing the
+            // player shouldn't also fix their hardsuit.
+            // TODO: The player drinking acid shouldn't deal heat damage to worn items. However, DamageSystem doesn't
+            // currently support this kind of modeling.
+            var dp = DamageSpecifier.GetPositive(damage);
```
EmoGarbage404, line 201: goes for the stuff in barotrauma system too, but modifications to upstream files should be enclosed with `// ES START` and `// ES END` comments for tracking where changes come. It just helps with resolving any potential merge conflicts.

EmoGarbage404 (changes requested):
Something i also considered was potentially having the puncturing be more granular: so perhaps having low amounts of damage not completely negate the pressure protection but instead reduce its efficacy until it breaks.

Content.Shared/Damage/Systems/DamageableSystem.cs
Partmedia: Sure.

Content.Server/Atmos/EntitySystems/BarotraumaSystem.cs
Partmedia: Sure.

Partmedia:
> Something i also considered was potentially having the puncturing be more granular: so perhaps having low amounts of damage not completely negate the pressure protection but instead reduce its efficacy until it breaks.

Sure, code-wise this is a pretty trivial change.

But seeing that the end goal is "pressurization simulation":

1. The suit doesn't enclose a lot of volume (suit volume minus body volume, leaving only a tiny volume between)
2. Even with a small hole, a small volume doesn't take a long time to equalize
3. So even in the end a small hole will result in the pressure equalizing rather quickly, i.e. no pressure protection

That, combined with the fact that there will be a lot more numbers to tweak, seems a little micro-balance-y to me. So please let me know.

Content.Server/Atmos/EntitySystems/BarotraumaSystem.cs
Partmedia: `PressureProtectionComponent` is currently server only. Not sure I want to be sharing and networking components at this stage.

EmoGarbage404: very annoying but understandable

PR #57, opened 2025-07-28
Ushankas

EmoGarbage404:
## About the PR

Adds ushankas from vg for use by command on russian station. I think they're a nice counterpart to the traditional beret so I wanted to port some nicer sprites than the goofy femboy-coded tg one.

## Media

image

**Changelog**


K-Dynamic:
Clothing using side sprites for icons is kinda cursed

Also idk if it matters downstream but meta.json is indented by two spaces instead of four

PR #78, opened 2025-08-15
Space bear sprite reverts

Flareguy:
## About the PR
Reverts the space bear sprites to their original versions. Also kills the point light associated with space mobs since it definitely doesn't make any sense with bright space + the space bear sprite.

Does not touch the other space mobs since those will probably be killed with hammers later.

## Why / Balance
I just don't really like the stylistic choice of having the space animals have this weird energy aura. In addition to looking ugly (it's a case example of bad unshaded layer usage,) space bears in of themselves are just meant to be an isolated, lightly space-themed gag mob. It would be really weird to have them in the russian station or what have you with their distracting glowing layers.
Additionally, the sprites just don't really fit the artstyle as well. They have way too much going on with their shading, and the simplistic, goofy construction of the old sprite fits in with other mob sprites and such a lot better IMO.

## Media
image

## Requirements

## Breaking changes

**Changelog**


mirrorcult:
sovl

UbaserB:
Why are they so adorable

EmoGarbage404:
not that i disagree, but we probably need the normal un-space bear sprites for the soviet station if we don't already have those

PR #91, opened 2025-08-18
Don't play gravity shake or powered light turn-on if round just started

mirrorcult:
this isnt a particularly clean solution but it works fine and i couldn't think of another one that wouldn't involve convoluted init flow shit


EmoGarbage404:
my hatred for the screen shaking outweighs my logical analysis of code

PR #92, opened 2025-08-18
Remove black spots baked into parallax background

mirrorcult:
it looks really bad. idk what flareguy is smoking

before:
image
image

after:
image


EmoGarbage404:
kinda agree but maybe my eyes are cooked because i straight up did not notice

mirrorcult:
it was easy to miss in gameplay but on the main menu it was rly obvious cuz it was part of the background and got scaled up way harder than everything else i think

PR #106, opened 2025-08-26
Relation icons

K-Dynamic:
## About the PR

- Adds good, neutral, bad relationship icons
  - Does not handle assigning icons to players for each relation
 
## Why / Balance

Useful for players to see relations of other characters at a glance, though they'll still need to examine to see exact relations.

Note for use: these should only appear if you have relations at all, otherwise display nothing. Otherwise you'll get UI spam and see everyone as 'neutral' despite not being related to them.

## Media

image

## Requirements

**Changelog**

Not player facing.


K-Dynamic:
I took a look at relationships.yml and I'm not sure if these three faces exactly fit (Rival would be considered neutral, Ex could be good/neutral/bad, etc.)

I think a good approach would be to let the player set the relationship icon with another player in their status menu, which effectively bookmarks other players.

EmoGarbage404:
either needs a plan and/or an implementation otherwise i dont see a point in just having 3 little face icons that im unsure will be used.

K-Dynamic:
> either needs a plan and/or an implementation otherwise i dont see a point in just having 3 little face icons that im unsure will be used.

I'll come up with something more specific for each of the current relationships.

PR #114, opened 2025-09-02
Make hv/mv electrocute through insuls when powered

EmoGarbage404:
## About the PR

Makes MV/HV wires electrocute through insulated gloves when cut. The only way to cut them is to disable power at the source (SMES, substation, generator, etc.)

## Why / Balance

This makes wire sabo more difficult since it limits mass snipping to only being possible on the APC/LV level. for any greater destruction, you need to break into a substation and attack it from there, which poses a greater barrier.

## Requirements

## Breaking changes

**Changelog**


ArtisticRoomba (approved):
its beautiful

PR #116, opened 2025-09-03
Adjust throw spin behavior + more tile friction

mirrorcult:
## About the PR

the bigger stuff
- throwing now calculates an integer amount of times to 'spin' the item based on distance (stepping by 1 every 4 tiles and starting at 0 spins)
- throwing resets angularvelocity/localrotation to 0 on landing, reorienting items (looks fine bc it should already be close to 0 rotation because of above)

minor changes
- throw speed slightly reduced
- tile friction increased by default, things stop faster now

motivation is to avoid having items be arbitrarily rotated often for readability purposes. throwing was 99.9% of the causes of that

https://github.com/user-attachments/assets/beba845d-1f84-4dfe-a9b5-f511955cc174


Sadie-silly:
holy peak

PR #117, opened 2025-09-05
Wallmount conditional visibility

mirrorcult:
## About the PR

https://github.com/user-attachments/assets/50c1cfb9-1c09-4e23-9bf5-91f4efb6d40e

- dont show wallmounts if youre on the wrong side
- show wallmounts always if your eye isnt drawing fov (i.e. ghost or something)
- works with arbitrary eye rotation/grid rotation/viewport rotation/what have you
- works with different viewports (!!)
- componenttree so it should be fast to query wallmounts in bounds

because we're enumerating wallmounts clientside, I ended up noticing that *windows and firelocks* have `WallMount` which is pretty bad. the only reason they do is so `InteractionSystem` can ignore anchored entities in their tile for interaction purposes, so I just made a separate comp with just that functionality and removed `WallMount` from them. dont have a video but they still work as expected.


mirrorcult:
cameras are always showing up as not visible for some reason, need to see what theyre doing

mirrorcult:
presumably its just cuz their sprites are actually the other way around from what is expected

K-Dynamic:
Interesting

mirrorcult:
> cameras are always showing up as not visible for some reason, need to see what theyre doing

just opted to make them arc 360 since theyre not actually placed on the wall anyway

Content.Client/_ES/Wallmount/ESWallmountVisibilitySystem.cs
```diff
@@ -0,0 +1,83 @@
...
+public sealed class ESWallmountVisibilitySystem : EntitySystem
+{
+    [Dependency] private readonly IEyeManager _eye = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly IPlayerManager _player = default!;
+    [Dependency] private readonly TransformSystem _xform = default!;
+    [Dependency] private readonly SpriteSystem _sprite = default!;
+
+    /// <inheritdoc/>
+    public override void Update(float frameTime)
```
moonheart08: Should be in FrameUpdate as part of normal rendering to avoid potentially popping in.

FrameUpdate is not predicted,so the prediction check can be removed.

*Alternatively*, it is entirely possible to do all this work in an overlay (maybe just call the function on the system to do the work so you can keep all the nice helpers), as rendering is single-threaded with the game itself, which would allow this to work with other viewports. You'd want to use a viewport underlay (ScreenSpaceBelowWorld or WorldSpaceBelowWorld) for this so it's run before all sprite rendering.

moonheart08 (changes requested):

Content.Client/_ES/Wallmount/ESWallmountVisibilitySystem.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        var playerXform = Transform(_player.LocalEntity.Value);
+        if (playerXform.MapID != _eye.CurrentEye.Position.MapId)
+            return;
+
+        var playerCoords = _eye.MapToScreen(_xform.GetMapCoordinates(playerXform));
+        if (!playerCoords.IsValid)
+            return;
+
+        var enumerator = AllEntityQuery<WallMountComponent, SpriteComponent, TransformComponent>();
```
moonheart08: This is a performance concern in replays. You probably want to set up a component tree for this so you can do accelerated lookups of what is in the viewport bounds. (which should be ~par for perf in normal gameplay, and notably better in replays)

mirrorcult: yeah, probably should have just used a componenttree from the start, shouldnt be that hard to do

Content.Client/_ES/Wallmount/ESWallmountVisibilitySystem.cs
mirrorcult: i deliberately didnt do frameupdate just bc i felt like there wasnt much risk in weird effects if it ran per tick instead of per frame (it looks fine, popin or whatever isnt noticeable) but i guess there isnt really a strict reason -not- to just have it in frameupdate

i dont think having this in an overlay makes sense, we aren't going to be able to render it per-viewport in a way that makes sense anyway because we have no way of changing how an entity is rendered depending on which viewport it's being rendered to/eye thats seeing it

moonheart08: Hm? If it's in an overlay it can just change the sprite visibility based on the eye being used for the currently drawing viewport. It'll rerun ahead of sprites being drawn in every viewport if put in ScreenSpaceBelowWorld.

So you effectively *can* change how the entities are being rendered.

mirrorcult:
reworked to use a componenttree + turned into an overlay, it now actually -does- work with arbitrary viewports

PR #119, opened 2025-09-07
Antimatter Engine

EmoGarbage404:
## About the PR

Overhauls the AME with a bunch of new mechanics and other stuff.

The AME is now based around _antimatter_ and _negentropy accumulators_. Antimatter is a voidlike substance that consumes matter and gases. As it consumes plasma gas, it grows, and as it consumes normal matter, it shrinks. Note however, it it fills the container, the antimatter will begin to explode violently.

The Negentropy accumulators harvest antimatter and convert it into energy. They can be damaged, but are easy to repair via welding tools.

## Media

ame-chamber

image

Content Client_FMg8odIuGn

## Breaking changes

**Changelog**


Content.Client/_ES/Power/Antimatter/ESAntimatterOverlay.cs
```diff
@@ -0,0 +1,112 @@
...
+    [Dependency] private readonly IEntityManager _entManager = default!;
+    [Dependency] private readonly IPrototypeManager _proto = default!;
+    [Dependency] private readonly IRobustRandom _random = default!;
+
+    private readonly HashSet<Entity<ESAntimatterComponent>> _antimatterSet = new();
+
+    public override OverlaySpace Space => OverlaySpace.WorldSpace;
+
+    private IRenderTexture? _aoTarget;
+    private IRenderTexture? _aoBlurBuffer;
```
mirrorcult: i wonder what u copied this from

Resources/Prototypes/_ES/Entities/Structures/Power/antimatter.yml
```diff
@@ -0,0 +1,193 @@
...
+        sound:
+          collection: MetalBreak
+      - !type:DoActsBehavior
+        acts: ["Breakage"]
+
+- type: entity
+  parent: BaseStructure
+  id: ESAntimatterCapsule
+  name: antimatter capsule
+  description: A highly-experimental container used for transporting antimatter. Safe enough, short of throwing explosives at it. But you'd never do that...
```
mirrorcult: seems weird for this to be the description when the intent is that you -should- blow it up

mirrorcult: overall this fact doesnt seem communicated very well (i know it doesnt have guidebooks yet but)

Resources/Prototypes/_ES/Entities/Structures/Power/antimatter.yml
```diff
@@ -0,0 +1,193 @@
...
+            max: 1
+      - !type:SpawnGasBehavior
+        gasMixture:
+          volume: 1000
+          moles:
+          - 0 # oxygen
+          - 0 # nitrogen
+          - 0 # carbon dioxide
+          - 100 # plasma
+          temperature: 373.15
```
mirrorcult, line 173: shouldnt the plasma be pumped in from outside to start it? didnt realize the can would have plasma in it too, is the intent that its like a bisected container

Resources/Textures/_ES/Structures/Power/negentropy.rsi/meta.json
```diff
@@ -0,0 +1,42 @@
+{
+    "version": 1,
+    "license": "CC0-1.0",
+    "copyright": "Created by EmoGarbage404",
```
mirrorcult: didnt flareguy make these

mirrorcult (commented):
for the overlay you probably want to look at whatever pjb is doing in https://github.com/space-wizards/space-station-14/pull/40181, ao was mentioned as being really inefficient so i assume this will have the same issue. though i dont think the engine commit that makes those work is versioned yet

mirrorcult:
also does the antimatter mass decay over time? i assume thats part of eventhorizon im just not seeing, but if not then it should get added. pretty necessary for gameplay otherwise u can just hook it up with a bit of plasma and leave it alone

EmoGarbage404:
> also does the antimatter mass decay over time? i assume thats part of eventhorizon im just not seeing, but if not then it should get added. pretty necessary for gameplay otherwise u can just hook it up with a bit of plasma and leave it alone

the mass decays when the accumulators make energy. Otherwise, it just stays constant. I didn't want it to naturally decay over time as I think it'd detract from the sorta "toxic waste" vibe it gets when it breaches containment and the cleanup gameplay you get from that.

mirrorcult:
ah that makes more sense, thats fine

Resources/Prototypes/_ES/Entities/Structures/Power/antimatter.yml
EmoGarbage404: It's just meant to give it a little bit more mass when it starts, but i changed the sprite to reflect that better.

EmoGarbage404:
someone remind me in however many weeks to update this with the overlay changes once engine is updated.

PR #124, opened 2025-09-10
Space Suits and Suit Storage

EmoGarbage404:
## About the PR

Adds a new basic EVA suit as well as a new line of suit storages designed to contain them.

The unique thing about the suit storage here is that it functions like a cabinet, having slots for the suit, helmet, tank, and mask. It also has visuals for the different items that can be placed inside of it (and i'll expand them when new suits and whatnot get added)

## Media

image
image

**Changelog**


Content.Shared/_ES/Core/Range/ESIntRange.cs
```diff
@@ -0,0 +1,30 @@
+using Robust.Shared.Serialization;
+
+namespace Content.Shared._ES.Core.Range;
+
+/// <summary>
+/// Defines a integer range, inclusive on both ends.
+/// </summary>
+[DataDefinition, Serializable, NetSerializable]
+public partial struct ESIntRange
```
moonheart08, line 9: Is https://learn.microsoft.com/en-us/dotnet/api/system.range?view=net-9.0 not serializable and thus usable here?

Resources/Prototypes/_ES/Entities/Structures/Storage/suit_storage.yml
```diff
@@ -0,0 +1,135 @@
...
+        state: fill-suit-base
+      "helmet":
+      - containerId: helmet_slot # no whitelist for base fill state
+        state: fill-helmet-base
+      "tank":
+      - containerId: tank_slot # no whitelist for base fill state
+        state: fill-tank-o2
+      "mask":
+      - containerId: mask_slot # no whitelist for base fill state
+        state: fill-mask-base
```
moonheart08: quotes + not intended makes this a little odd to read.
either indent further or remove the quotes

(nitpick)

Content.Shared/_ES/Storage/ItemMapper/ESSharedItemMapperSystem.cs
```diff
@@ -0,0 +1,74 @@
...
+            layers.Add(layerKey, layerState);
+        }
+
+        Appearance.SetData(ent, ESItemMapperVisuals.Layers, layers);
+    }
+
+    private bool IsMappingSatisfied(Entity<ESItemMapperComponent, ContainerManagerComponent> ent, ESItemLayerMapping mapping)
+    {
+        if (!_container.TryGetContainer(ent, mapping.ContainerId, out var container, ent))
+            return false;
```
moonheart08: This should be an error condition, not a simple return. If no `GetContainer` exists, then 
```csharp
throw new Exception($"Couldn't find the container {mapping.ContainerId} for mapping layer {layerKey} on {ToPrettyString(ent)}.")
```

Content.Shared/_ES/Storage/ItemMapper/ESSharedItemMapperSystem.cs
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private void OnEntRemoved(Entity<ESItemMapperComponent> ent, ref EntRemovedFromContainerMessage args)
+    {
+        UpdateMappings(ent);
+    }
+
+    public void UpdateMappings(Entity<ESItemMapperComponent> ent)
+    {
+        if (!TryComp<ContainerManagerComponent>(ent, out var containerManager))
```
moonheart08: This should be an error condition, not a simple return. Otherwise you're gonna discover container typos at runtime while trying to play the game instead of during map or object init.

Content.Shared/_ES/Storage/Slots/ESOpenableSlotSystem.cs
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private void OnClosed(Entity<ESOpenableSlotsComponent> ent, ref OpenableClosedEvent args)
+    {
+        UpdateSlotsLocked(ent);
+    }
+
+    private void UpdateSlotsLocked(Entity<ESOpenableSlotsComponent> ent)
+    {
+        if (!TryComp<ItemSlotsComponent>(ent, out var slots))
```
moonheart08: This should be an error condition, not a simple return.

Content.Client/_ES/Storage/ItemMapper/ESItemMapperSystem.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        SubscribeLocalEvent<ESItemMapperComponent, AppearanceChangeEvent>(OnAppearanceChange);
+    }
+
+    private void OnAppearanceChange(Entity<ESItemMapperComponent> ent, ref AppearanceChangeEvent args)
+    {
+        if (args.Sprite is not { } sprite)
+            return;
+
+        if (!Appearance.TryGetData(ent, ESItemMapperVisuals.Layers, out Dictionary<string, string?> layers, args.Component))
```
moonheart08, line 24: Ditto, this should be an error condition given this component's sole purpose is visuals and it will silently break without the layers.

```csharp
throw new Exception($"Couldn't find the necessary {nameof(ESItemMapperVisuals.Layer)} layer on {ToPrettyString(ent)}.");
```

Content.Client/_ES/Storage/ItemMapper/ESItemMapperSystem.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<ESItemMapperComponent, AppearanceChangeEvent>(OnAppearanceChange);
+    }
+
+    private void OnAppearanceChange(Entity<ESItemMapperComponent> ent, ref AppearanceChangeEvent args)
+    {
+        if (args.Sprite is not { } sprite)
+            return;
```
moonheart08: Not ditto here because I'm not entirely sure if this is affected by things like visibility... if it's not, then this is probably also an error condition.

moonheart08 (changes requested):
Visuals look good. I don't have much of any comments on the appearance code aside from the error handling, it seems reasonably written and it's not my strong-suit.

moonheart08 (approved):

PR #129, opened 2025-09-11
Old smartfridge sprites

Flareguy:
## About the PR
Adds back the old TG smartfridge sprites

## Media
Content Client_mJQ8bNtYon

## Requirements

## Breaking changes

**Changelog**


UbaserB:
yummy sprites

EmoGarbage404 (changes requested):
The booze-o-mat should definitely be a wallmount. That is almost certainly how we want it to be used in-game so having it just be a full-tile vendor would feel pretty bad.

As for the fridge, while I'm fine with the sprite, if we plan on using it as a literal wall and it has the same perspective as walls, then it should at least block air, not be trivially unwrenchable, etc.

moonheart08:
> The booze-o-mat should definitely be a wallmount. That is almost certainly how we want it to be used in-game so having it just be a full-tile vendor would feel pretty bad.
> 
> As for the fridge, while I'm fine with the sprite, if we plan on using it as a literal wall and it has the same perspective as walls, then it should at least block air, not be trivially unwrenchable, etc.

do the based thing and make the boozeomat directional with smoothing?
This would probably require somehow making only the display a wallmount while preserving its full-tile status.

..maybe just make it an actual wallmount lol.

PR #134, opened 2025-09-12
Window tranparency tweaks and remove xeno+uranium windows

UbaserB:
## About the PR

Makes the inner section on windows more transparent (except for the shuttle+plastitanium ones), and removes the xeno+uranium windows. I removed them because they're ugly and bloat the list, while the transparent window tweaks feel more sovl and defined rather than a flat blob of color.

## Media

image

## Requirements


UbaserB:
forgot to fix something...

PR #137, opened 2025-09-14
Clientside inherent mob lights

mirrorcult:
image

this took me like 2.5 hours instead of 20 minutes because spawning the entity and doing `SetParent` didnt make the light show up but using `SpawnAttachedTo` did make the light show up

works entirely clientside on attach/detach. no serverside entities spawned for this. cleans itself up on detach so there's only ever one

should probably add different prototypes for mobs that can see in the dark better (like spiders or something)


mirrorcult:
image
awesome

PR #138, opened 2025-09-14
Pull tweaks

mirrorcult:
## About the PR

- fixedrotation for items being pulled (avoids arbitrary rotations from items being pulled against corners)
- no inherent slowdown from pull. mass slowdown still exists of course which i want to make less prominent but that will require a more involved solution
- shorter pull range and pull max length. just looks better imo, also the maxlength is actually consistent instead of being based on where you pulled it from which was always really weird to me. i might even want it lower (although keep the pull range at 1 tile)


mirrorcult:
image
this fucking test

mirrorcult:
image
awesome

mirrorcult:
pulling unbuckling people obviously still works so this is just a sloptest whatever i dont wanna think about it rn

mirrorcult:
actually the testfail is probably from interactiontests spawning the target farther away than expected

mirrorcult:
fuck my gay life

mirrorcult:
gay test success

PR #141, opened 2025-09-16
Move old ghost light system to inherent lights + disable by default for ghosts

mirrorcult:
## About the PR

image

also api for enabling/disabling it in the future


mirrorcult:
actually im just gonna. go the whole way and replace the entire thing here

mirrorcult:
image

mirrorcult:
image
this could truly not be further from being a networked entity

PR #152, opened 2025-09-20
Audit + tweak component property animations pass 1

mirrorcult:
## About the PR

uses new animation track easing shit i added to engine

main meat is melee stuff. rest is just kinda obvious shit
- mob lunge is now same as tg and actually looks good
- fixes animations for shit to actually use wideanimation so spear thrust et c looks correct
- makes swings actually swings + tweaks length and shit for all them and it looks good nw + easing and all that rubbish
- fix the fade on melee effects it was just straight broken before
- made swings alternate swing direction instead of just being from the same direction every time
- it still doesnt look amazing but its a lot better i think

other non obvious things
- pointy bouncy
- pickup anim now scales the item down also so it looks like its. going to you and getting shrinky so it goes into your. inventory.

point (slight out dated)

https://github.com/user-attachments/assets/c65cbd67-a179-4274-b477-6b2b0ad2ca2d

pickup

https://github.com/user-attachments/assets/310a9952-d0e2-4165-83e1-f8e3cafba4cd

throw

https://github.com/user-attachments/assets/ffc8a137-f277-4678-9208-0dd028e932dd

melee stuff

https://github.com/user-attachments/assets/cc2d1a9f-3010-4b2f-a33c-7947e2ba75de

### DIFF
--- a/Content.Client/Animations/EntityPickupAnimationSystem.cs
+++ b/Content.Client/Animations/EntityPickupAnimationSystem.cs
@@ -68,7 +68,7 @@ public sealed class EntityPickupAnimationSystem : EntitySystem
 
         _animations.Play(new Entity<AnimationPlayerComponent>(animatableClone, animations), new Animation
         {
-            Length = TimeSpan.FromMilliseconds(125),
+            Length = TimeSpan.FromMilliseconds(300),
             AnimationTracks =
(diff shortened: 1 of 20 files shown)


UbaserB:
so peak...

PR #164, opened 2025-09-26
the cone man

mirrorcult:
## About the PR

ports bhijn's og viewcone patch, cleans it up significantly, adds more stuff onto it, makes it work better in more situations

various things in a bullet point list below
- item alpha modulation happens in 2 overlays, one that runs before entities to set alpha and one that runs after entities to reset alpha back to what it was. so, alpha is never changed at any other point of the tick, meaning this effect works pretty much gracefully with other things that set alpha
- uses a componenttree to query for fadeable/occludeable entities for perf
- works multiviewport, supports stuff like cameras having viewcones
- supports 'viewcone effects', aka just entities that have an inverted fade and only show up when out of view, also sanely works with stuff like footsteps where it wont show your own footsteps but will show other peoples
- defaults mouserotator to being on always (not just combat mode) Teehee my grand plan
- finagles esbasemob stuff to not always be part of basemob itself so that i can escape hell and stop giving observers viewcones

things that are not in this pr and should be later
- i only added footstep effects, nothing else, but there are sprites for attack/gun/etc effects which would be good to have
- i thiiiiink this can mildly look wrong with some parent situations but can be fixed by just adding viewconeoccludeable to stuff i missed i thiiiink
- no support for equipment modifying viewcone. ik how i want to implement this but im tired and ill do it later

the video compression kind of fucks it but there is like a subtle staticy noise in the viewcone. i think its pretty subtle and not jarring or anything but its hard to show what it actually looks like. tr=ust me its like fine

video showing the general shits

https://github.com/user-attachments/assets/04df2fae-b27a-4deb-9970-aa1475043926

video showing effects

https://github.com/user-attachments/assets/4cab1151-d99e-4acc-a057-dff89b998f40

camera that i added a viewcone to manually + multiviewport etc

image


Content.Client/_ES/Viewcone/Overlays/ESViewconeSetAlphaOverlay.cs
```diff
@@ -0,0 +1,129 @@
...
+        while (enumerator.MoveNext(out var uid, out var eye, out var viewcone))
+        {
+            if (args.Viewport.Eye != eye.Eye)
+                continue;
+
+            _nextEye = (uid, eye, viewcone);
+            break;
+        }
+
+        return _nextEye != null;
```
mirrorcult, line 62: realistically the most sus thing here. couldnt think of a better way to go from an `IEye` to the `EyeComponent` that holds it without engine changes. but also like this is completely fine

Resources/Prototypes/_ES/Entities/Effects/viewcone_effects.yml
```diff
@@ -0,0 +1,74 @@
...
+  - type: ESViewconeOccludable
+    inverted: true
+  - type: Sprite
+    sprite: _ES/Effects/viewcone_effects.rsi
+    drawdepth: Effects
+    noRot: true
+    color: '#FFFFFF00' # 0 alpha by default, will get overriden by setalpha overlay
+    # what the fuck? removing looping didnt exist until literally a few hours ago?
+    # and now i have to version engine again?
+    # loop: false
```
mirrorcult: this makes the effects look kinda wrong unless i make them despawn faster than they should since they just loop automatically lol. but once we update engine ill fix

mirrorcult:
image
possible alternate impl for how viewcone moves i might try out. i agree with some concerns that the way it works rn could end up looking really stupid in actual gameplay. will try implementing this and seeing how it feels

EmoGarbage404:
i believe in the cone man

PR #166, opened 2025-09-26
Numerous gun additions and changes

Flareguy:
## About the PR
Adds ES versions of the shotgun, double-barrel shotgun, hybrid taser, and energy rifle.
Migrates out some gun-related stuff that has been replaced as well.

## Media
Content Client_vUVXInRJSL

## Requirements

## Breaking changes

**Changelog**


Sadie-silly:
the double barrel looks really short, like it's sawn off almost.
the broken sprite also looks like the barrel randomly gets longer which looks really weird

Flareguy:
Things that are currently broken / would really benefit from added functionality:
- Adding a fixture to any projectile in bullets.yml crashes the game for no reason, even when the YAML is directly copy-pasted from existing projectiles.
- BatteryWeaponFireModes needs to support being able to change sprites based on firing mode
- BatteryWeaponFireModes needs to support being able to change firing sound based on firing mode

Stuff that isn't needed but would be really nice:
- There should probably be a way to have a pointlight based on the charge level of a gun (to combat ugly overcontrasting unshaded lights)
- Having storage fill visuals that work like MagazineVisuals and automatically fill without having to make a million states on the sprite itself would save me a lot of time.

PR #175, opened 2025-09-29
Backwards walking + move walk to space

mirrorcult:
## About the PR

u now walk when moving backwards (using strafe key/combat mode mouserotator)

also moves walk to space by default to not conflict with strafe

image

https://github.com/user-attachments/assets/2563a025-9173-4e69-be80-e352774bb53a

now you must be Scared while running because of the Cone man is also there


mirrorcult:
wait no merge yet 1 more thing

mirrorcult:
ok now you also no footsteps on walk. this incldues the viewcone effect. sneak around

Content.Shared/Movement/Systems/SharedMoverController.Input.cs
```diff
@@ -373,15 +373,17 @@ private void HandleRunChange(EntityUid uid, ushort subTick, bool walking)
             SetSprinting((uid, moverComp), subTick, walking);
         }
 
-        public (Vector2 Walking, Vector2 Sprinting) GetVelocityInput(InputMoverComponent mover)
+        // ES START
+        // forceWalk
+        public (Vector2 Walking, Vector2 Sprinting) GetVelocityInput(InputMoverComponent mover, bool forceWalk=false)
```
moonheart08, line 380: Effectively an upstream's problem nitpick: This function probably shouldn't be in charge of whether you're walking or not, and the idea of a character having two separate velocity vectors is a bit of a code smell.

Content.Shared/Movement/Systems/SharedMoverController.Input.cs
```diff
@@ -373,15 +373,17 @@ private void HandleRunChange(EntityUid uid, ushort subTick, bool walking)
             SetSprinting((uid, moverComp), subTick, walking);
         }
 
-        public (Vector2 Walking, Vector2 Sprinting) GetVelocityInput(InputMoverComponent mover)
+        // ES START
+        // forceWalk
```
moonheart08: this should probably be more descriptive of the changes being done, think of it as an explainer for whoever has to fix it when a merge conflict happens.

Content.Shared/Movement/Systems/SharedMoverController.cs
```diff
@@ -274,7 +279,16 @@ protected void HandleMobMovement(
...
+            // ES START
+            var backwardsAngle = moveSpeedComponent?.BackwardsAngle ?? MovementSpeedModifierComponent.ESDefaultBackwardsAngle;
+            var rotNorm = worldRot.ToWorldVec().Normalized();
+            var velNorm = velocity.Normalized();
+            var cosAngle = Vector2.Dot(velNorm, rotNorm);
+            var threshold = new Angle(MathF.PI) - (backwardsAngle / 2);
+
+            forceWalk = cosAngle < Math.Cos(threshold.Theta);
+            wishDir = AssertValidWish(mover, walkSpeed, sprintSpeed, forceWalk);
+            // ES END
```
moonheart08, line 296: Clarity of scope nitpick, using a block to put all these variables in a neat bundle so its clear when they're used.

```suggestion
            // ES START
            // Check if we're walking backwards within the given angle range, and if so force walking speed.
            {
                var backwardsAngle = moveSpeedComponent?.BackwardsAngle ?? MovementSpeedModifierComponent.ESDefaultBackwardsAngle;
                var rotNorm = worldRot.ToWorldVec().Normalized();
                var velNorm = velocity.Normalized();
                var cosAngle = Vector2.Dot(velNorm, rotNorm);
                var threshold = new Angle(MathF.PI) - (backwardsAngle / 2);
  
                forceWalk = cosAngle < Math.Cos(threshold.Theta);
                wishDir = AssertValidWish(mover, walkSpeed, sprintSpeed, forceWalk);
            }
            // ES END
```

Resources/keybinds.yml
```diff
@@ -48,7 +48,9 @@ binds:
   key: S
 - function: Walk
   type: State
-  key: Shift
+# ES START
+  key: Space
```
moonheart08, line 52: should this really be rebound? Most players likely have shift-walk muscle memory, even if I do think this is the better keybind. Maybe put strafe here instead?

Content.Shared/Movement/Systems/SharedMoverController.Input.cs
mirrorcult: yeah its really awful to work with and i had a shelved commit changing a lot more of it but decided it wasnt worth it

Resources/keybinds.yml
mirrorcult: it works weirdly with strafing if theyre on the same key. id rather just have it be the better bind and explain it to people

PR #184, opened 2025-10-04
NT Toast

EmoGarbage404:
## About the PR

NT Toast, by Emisse.

Thanks to flareguy with additional help mapping atmos.

## Media

ESToast-0

## Breaking changes

**Changelog**


UbaserB:
sustenance vendor is the wrong vendor
also theatre drobe

mirrorcult:
should toast get set as the default map instead of testmap?

PR #225, opened 2025-10-19
Add secondary votes to events

EmoGarbage404:
## About the PR

Events now can have secondary votes that determine aspects about them. adds in custom logic for handling all this in a nice way.

only has implementation for gas leak rn

## Media

image

### DIFF
--- a/Content.Shared/_ES/Voting/Components/ESSynchronizedVoteManagerComponent.cs
+++ b/Content.Shared/_ES/Voting/Components/ESSynchronizedVoteManagerComponent.cs
@@ -1,0 +1,65 @@
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using Robust.Shared.GameStates;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Utility;
+
+namespace Content.Shared._ES.Voting.Components;
+
+/// <summary>
+/// This is used for an entity which spawns multiple child <see cref="ESVoteComponent"/> entities.
+/// These votes are run in tandem and, when completed,
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+[Access(typeof(ESSharedVoteSystem))]
+public sealed partial class ESSynchronizedVoteManagerComponent : Component
+{
+    [ViewVariables]
+    public bool Completed => !Results.Any(r => r is null);
+
+    /// <summary>
+    /// The vote prototypes that will be spawned
+    /// </summary>
+    [DataField]
+    public List<EntProtoId<ESVoteComponent>> Votes = new();
+
+    /// <summary>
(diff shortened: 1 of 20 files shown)


Content.Shared/_ES/Voting/Components/ESSynchronizedVoteManagerComponent.cs
```diff
@@ -0,0 +1,65 @@
...
+using Robust.Shared.GameStates;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Utility;
+
+namespace Content.Shared._ES.Voting.Components;
+
+/// <summary>
+/// This is used for an entity which spawns multiple child <see cref="ESVoteComponent"/> entities.
+/// These votes are run in tandem and, when completed,
+/// </summary>
```
mirrorcult, line 12: ,

Resources/Locale/en-US/_ES/station-events/gas-leak.ftl
```diff
@@ -0,0 +1,2 @@
+es-voter-query-string-gas-leak-gas = The gas that will be leaked is:
+es-voter-query-string-gas-leak-location = The gas will be leaked:
```
mirrorcult:
```suggestion
es-voter-query-string-gas-leak-gas = The gas that should be leaked is:
es-voter-query-string-gas-leak-location = The gas should be leaked:
```

"will" kind of makes it seem like youre trying to guess where it'll go rather than influencing where it goes

Resources/Prototypes/_ES/GameRules/station_events.yml
```diff
@@ -40,7 +43,37 @@
...
+    - Ammonia
+    - Plasma
+    - Tritium
+    - Frezon
+    - WaterVapor # The Fog...
+    count: 2
+
+- type: entity
+  id: ESVoteGasLeakLocation
+  name: Where Will the Gas Leak?
```
mirrorcult:
```suggestion
  name: Where should the gas leak?
```

Resources/Prototypes/_ES/GameRules/station_events.yml
```diff
@@ -40,7 +43,37 @@
...
     endAnnouncement: station-event-gas-leak-end-announcement
   - type: ESGasLeakRule
+  - type: ESSynchronizedVoteManager
+    votes:
+    - ESVoteGasLeakGases
+    - ESVoteGasLeakLocation
+
+- type: entity
+  id: ESVoteGasLeakGases
+  name: Which Gas Will Leak?
```
mirrorcult:
```suggestion
  name: Which gas should leak?
```

are these supposed to be title cased. idk what the thought is there

mirrorcult (changes requested):

PR #240, opened 2025-10-30
Decal & white tile adjustments + toast fix

Flareguy:
## About the PR
Makes a few extremely minor adjustments to decals and white tiles. Decals now have slightly less alpha, and white tiles now have more compressed upper contrast bands.

Also updates toast to change the decal colors + fix some APC mapping errors while I'm at it.

## Why / Balance
These minor contrast adjustments should really help to make the scene's contrast level slightly more comfortable feeling. Even if they are hard to notice, I think they make a difference.

## Media
You probably can't tell the difference, but it looks better, okay?

Content Client_l6WM0oGbla

## Requirements

## Breaking changes

**Changelog**


EmoGarbage404:
This PR is psychotic just because, even upon close examination, the difference between the tiles is so minimal as to be barely perceptible.

PR #249, opened 2025-11-01
Directional posters, rejig lights, toast pass for both

mirrorcult:
## About the PR

no render because this doesnt really change anything visible to a render

i had to add a new poster comp in order to do directional posters without adding an insane amount of annoying yaml to every existing poster (it just sets a state idk now that i think about it this probably exists already)

added support for markers to only hide specific layers (used to show directionality on otherwise adirectional-sprite wallmounts), adds this to posters and extinguisher cabinets

adds new poster spawners to use for ES that actually use entitytables, we can change the pool later

remaps toast for poster directionality, used my best judgment in cases where it wasnt clear what direction a poster was 'supposed' to be facing, in some cases where it seemed like it should be in multiple i just added new posters

changes light values and remaps toast to have a better light budget (maxes out at 45 shadowcasters at any given time generally) also its just slightly less dim overall idk still feeling out how i want it to look i think

image
image

### DIFF
--- a/Content.Client/Markers/MarkerComponent.cs
+++ b/Content.Client/Markers/MarkerComponent.cs
@@ -7,5 +7,10 @@ namespace Content.Client.Markers
     [RegisterComponent]
     public sealed partial class MarkerComponent : Component
     {
+        // ES START
+        // used to only toggle specific layers rather than the entire sprite
+        [DataField]
+        public HashSet<string>? Layers;
+        // ES END
     }
 }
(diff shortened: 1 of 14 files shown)


EmoGarbage404:
test fail certified REAL

devicenetwork hanky

mirrorcult:
at some point i deleted an air alarm by accident and panicked and went to go replace it so its probably just that

mirrorcult:
test figs

PR #251, opened 2025-11-02
Do you like my wonderful space

mirrorcult:
## About the PR

regular parallax
image

with lores viewport/lq parallax on (would look bad with scaling otherwise)
image

scale down for vast ness of space and lonelyness

pitch blackness is scary scary space

denser dust field

darker black on parallax

2x res on star texture + denser and with some sillies on it

no random rotation on toast bc rotation uniquely makes this parallax looks kinda bad

change parallax to load lq verison on lores viewport for scaling reasons above. you can also just have the lq thing checked separately but i wanted to force it on lores viewport always cuz it looks bad otherwise


mirrorcult:
ok it real now

PR #264, opened 2025-11-11
APC-based powerhum + resaves 2 apcs on toast

mirrorcult:
## About the PR

- shipambience.ogg (power hum) on apcs
- ambientsound, added functionality to have a max of 1 for just that sound + play nonpositionally so it can be stereo and not feel weird
- inverts the importance calc for ambientsound range, now it considers closer ambient sources to be better. this doesnt matter except for cases of max sounds
it doesnt fade out very good rn but its fiiiine we dont have good fades in engine anyway, ill fix it when we get that
i like power hum bc it adds a layer to basic audio without being overbearing + the main draw is that you notice when its not playing (i.e. when power is out or when youre not near the station). feels very fun

resaves toast to move the tcomms apc for better power hum coverage (was the only couple tiles on the map i noticed where it was leaking) and to fix an apc near bridge, also fixgridatmos cuz i think something was broken

intentionally turned up a bit to show the effect

https://github.com/user-attachments/assets/982b3b6c-cb32-4e90-9a3b-f18a9c833647


mirrorcult:
https://github.com/user-attachments/assets/f4be4f22-7425-4bd3-8e3c-49769875e4ae

PR #281, opened 2025-11-16
Basic arrivals flow

mirrorcult:
## About the PR

map by flare. not perfect but gets the job done for now. eventally i want like cafeteria in the middle instead of seats area. you know like. luncheon table. type shit

map modified by me to moe the docking side to the other side cause it didnt fit on toast otherwise

how it fits
image

full flow w lobby
obviously ftl time is longer normally just reduced to make the point clear

https://github.com/user-attachments/assets/906e71eb-c3b3-4954-9849-b042277c9b76

things to do later
- actually start people in the cryosleep and have like a . gaspy effect when u get out. and chance to be all woozy. and be hungry and thorsty
- oneway doors somehow (turnstile maybe) on arrivals so people cant go back into the shuttle
- remap shuttle a bit to have more sleepers (theres only 40 right now sooo eh not a big issue i guess)
- maybe job-specific cryosleeps so security spawns near sec and cap spawns in cap room or bridge etc etc etc
- more sauce overall
- toast needs better signposting near arrivals. theres like no signs on this damn map


mirrorcult:
arrivals-0

render
idk why but it didnt render the random vendor spawns which u can see in the image in the pr body i guess

Content.Server/_ES/Arrivals/ESArrivalsSystem.cs
```diff
@@ -207,11 +123,13 @@ public void HandlePlayerSpawning(PlayerSpawningEvent ev)
...
             ev.Station);
 
-        EnsureComp<AutoOrientComponent>(ev.SpawnResult.Value);
+        // TODO MIRROR one-way arrivals, use these for a turnstile check or something + remove on exiting arrivals
+        // EnsureComp<AutoOrientComponent>(ev.SpawnResult.Value);
         var passenger = EnsureComp<ESArrivalsPassengerComponent>(ev.SpawnResult.Value);
         passenger.Station = ev.Station.Value;
     }
 
+    // TODO MIRROR why does this even exist
```
EmoGarbage404: originally mask stuff was based on player's arriving at the station rather than just roundstart  spawning, so this event was used to grab all the players who spawned and assign them masks. however, if arrivals wasn't enabled, then this event would never be raised and mask assignment would break locally.

it can probably be removed now. i don't really know what kind of purpose it would even serve.

Content.Server/_ES/Arrivals/ESArrivalsSystem.cs
```diff
@@ -99,104 +93,26 @@ private void OnShuttleTag(Entity<ESArrivalsShuttleComponent> ent, ref FTLTagEven
         args.Tag = DockTagProto;
     }
 
-    private void OnFTLStarted(Entity<ESArrivalsShuttleComponent> ent, ref FTLStartedEvent args)
```
EmoGarbage404, line 102: it may be prudent to keep this as i think having a countdown to the shuttle arrival on-station is a good thing. maybe just move it to the FlyToStation code

mirrorcult: oh yeah I guess I wasn't thinking about screens on the shuttle itself that show the countdown, I can readd it

mirrorcult:
image

screen stuff works again + added some to the shuttle

PR #288, opened 2025-11-18
Prune some names

mirrorcult:
## About the PR

mostly just names i looked at and thought 'i will never find this name interesting no matter what its next to' or were dupes or were just variations. or were just low quality (one of the gundam names being Ple was crazy)

recommend looking at https://github.com/EphemeralSpace/ephemeral-space/commit/532f7953cf17e6baa35c49f5a3dc5403babf12d2 for a cleaner diff


mirrorcult:
this will be final name adjusting pr for a while i dont want more

PR #290, opened 2025-11-20
Smart Internals

EmoGarbage404:
## About the PR

Proposed by mirrorcult, a more streamlined approach to managing internals.
One of the issues with internals right now is that you tend to pick up and drop tanks over time, shuffling around actions in the hotbar. Additionally, every tank needs its own action, which is unnecessary since you only ever use one tank at most. This significantly contributes to actions bloating up the screen, which is unwelcome.

the new system smartly infers how to connect to internals based on context. If you have no tank or mask, it'll inform you when you press the button. If you have one tank, then the connection will happen automatically with no further input. If there are multiple, when connecting, you'll be prompted to choose one via a radial menu.

https://github.com/user-attachments/assets/c04345eb-d2f2-4861-899e-74186bd797d8

image

### DIFF
--- a/Content.Client/Input/ContentContexts.cs
+++ b/Content.Client/Input/ContentContexts.cs
@@ -51,6 +51,7 @@ namespace Content.Client.Input
 
             // ES START
             common.AddFunction(ContentKeyFunctions.ESHoldToFace);
+            common.AddFunction(ContentKeyFunctions.ESToggleInternals);
             // ES END
 
             var human = contexts.GetContext("human");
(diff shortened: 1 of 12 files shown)


mirrorcult:
not sure how robust the radialui is but itd be nice for the tooltip on the radial elements to have a suffix for the inventory slot it comes from. like 'emergency oxygen tank (pocket1)' for disambiguations sake

EmoGarbage404:
image

disambiguated

PR #309, opened 2025-11-24
Parallax & maplight variation passes

mirrorcult:
## Why / Balance

had to tweak some variation stuff (add an optional entitytable variant lol)

makes the default maplight not #ffffff (looks cooler this wya)

adds command for debug testing variation passes

entity table logic for it is kind of convoluted but also beautiful and also not that hard to understand my reasoning i think

these are all kind of semi low effort. you can do a lot with more involved parallax stuff but i just wanted a simple proof of concept but still sauced version initially. id like to have more random objects appear in parallax but ill need to mess with the parallax system itself to do that. also bay has procgen planet parallaxes which look awesome. also i wanna think about random nebulas but id need good nebula textures first and ermmmm

xamples

hey whered the stars go
image

its looking awfully red out today
image

The light of heaven has been snuffed out
image


mirrorcult:
oh yeah this also adds an ES-specific variation pass table but ill do a proper pass over it later

PR #317, opened 2025-11-26
Latejoining curtains

### LINKED ISSUE #298: Latejoining doesn't play the curtain effect
i forgot

mirrorcult:
## About the PR

closes #298

added an entitytimer api for broadcast (saves to nullspace)

https://github.com/user-attachments/assets/f915ad52-4fb7-4c5d-8f88-f964e00cf2b5


mirrorcult:
also i changed the ease from quint to quad cause i think it looks better

PR #335, opened 2025-11-30
Blinking

EmoGarbage404:
## About the PR

I got bored.

TabTip_VCpQRD0L2M
image

humans blink every so often. when sleeping, they close their eyes and when dead they get locked open.


Emisse:
please add manual blinking trait that makes you have to manually blink or else your vision gets blurry please please please please please

EmoGarbage404:
Ed has reported that the eyes dont re-open after waking up so that'll need to be investigated.

EmoGarbage404:
wasn't able to replicate so this goes into the lived human experience

PR #337, opened 2025-12-01
Arms Dealer

EmoGarbage404:
## About the PR

Adds arms dealer according to [doc](https://ephemeralspace.github.io/docs/design/masks/crew/arms-dealer.html)

image

arms dealer can summon guns out of closets like a fruit vendor. their objective is to make sure people hold onto their guns (either in bag or in hand at end of round)

"no kills" objective not implemented for now because it's hard and mirror said it could be left out on the first pass.


Resources/Prototypes/_ES/Actions/Masks/arms-dealer.yml
```diff
@@ -0,0 +1,36 @@
...
+          weight: 5
+        - group:
+          - id: ESWeaponShotgun
+            weight: 2
+          - id: ESWeaponDoubleBarrelShotgun
+            weight: 1
+          weight: 3
+        - group:
+          - id: ESWeaponEnergyRifle
+          weight: 2
```
mirrorcult: IMO base arms dealer table should only include small arms and only ballistics. if it does include something larger (imo a sawn off shotgun would be fine) it should be weighted much lower than small arms

PR #341, opened 2025-12-03
mildly large toast update + an entire new lighting pass again

### LINKED ISSUE #136: Spacelight leaks through walls
Image

should be easy to fix by just making it propagate less. not sure where thats configured though

mirrorcult:
## About the PR

okay this is a lot of stuff in one and it doesnt include the main thing i wanted to do (resprite and map lightswitches everywhere) but here are the changes and why

**1)** lighting pass again. sorry. i really like how it looks this time for actual though and i dont think ill majorly change it again. summary is as follows
  **1.1)** general station lighting (fluorescents) has been split into two types. department lights (brighter, used in most Rooms) and hallway lights (dimmer, unresponsive to lightswitches). this is because i think it looks really good and it makes for a nice contrast between transitory areas and more 'inviting' brighter areas. much more fun this way. in general it fulfills my desire to have moody lighting while still having high visibility in areas that should have high visibility (departments)
  **1.2)** lightblur factor has been increased overall, this makes softshadows look much better in my opinion (especially with clashing light colors like i have introduced) and in general makes stuff look a little better. i had to reduce the multiplier in contents lightbluroverlay (spacelight and stuff) because otherwise it was untuned
  **1.3)** there is also a new maintenance smalllight variant (the only diff is it cant be triggered by lightswitches which is an obvious problem otherwise) but its not mapped yet ill do it when i actually map and test light switches

**2)** lots of changes to visibility/sightlines on toast which are relevant for lightleak reasons, stealthy sightline reasons and arrival vibe reasons
  **2.1)** all situations where rooms had a shutter control that only shut over tables have been changed to shut over nearby windows, thus actually blocking sightlines into a room
  **2.2)** more rooms have been given shutters over their windows, and in general these shutters are closed roundstart (again for arrivals vibes reasons see arrivals doc). this includes: engineering foyer, sec (bottom area, main sec is still easy to see into), cargo foyer, botany, det room (closed the blinds), security checkpoint (open roundstart), chemistry, rnd/robotics. notably, not main sec area, the kitchen, the bar, or medbay
  **2.3)** most rooms which had glass airlocks, but windows into the room, have been changed to be solid airlocks, again so that sightlines into the room can be completely shut off if desired. this includes botany, cargo foyer, the kitchen, but -not- sec or medbay

**3)** adds directional signage to toast. it didnt have this and it bothered me so i added some. for the most part, they are assuming the player is coming from arrivals (as they all will be) and so is catered around navigating from arrivals to the other departments, not necessarily from other departments to anywhere else. i didnt want to bloat walls too much, but they should still be very useful.

and also a veritable motley of other random stuff
- fixes #136 by accident mostly
- station maps are now directional
- introduces directional buttonframes, remaps some buttons to be nondirectional which were directional by accident (ones on tables and stuff)
- adds a recharger into the sec checkpoint. i am open to removing this but this was a staple in ss13 maps and it made these rooms a lot more useful than they otherwise would be and its just kind of something i noticed didnt exist

ok now screenshots click for fullres

### DIFF
--- a/Content.Client/Entry/EntryPoint.cs
+++ b/Content.Client/Entry/EntryPoint.cs
@@ -156,7 +156,7 @@ namespace Content.Client.Entry
             // we cant set engine cvars by changing the file,
             // and cant set client-only non-replicated cvars with configpresets
             // so
-            _configManager.SetCVar("light.blur_factor", 0.0037f);
+            _configManager.SetCVar("light.blur_factor", 0.0025f);
             // ES END
         }
 
--- a/Content.Client/Entry/EntryPoint.cs
+++ b/Content.Client/Entry/EntryPoint.cs
@@ -151,6 +151,13 @@ namespace Content.Client.Entry
             _configManager.SetCVar("interface.resolutionAutoScaleLowerCutoffX", 520);
             _configManager.SetCVar("interface.resolutionAutoScaleLowerCutoffY", 240);
             _configManager.SetCVar("interface.resolutionAutoScaleMinimum", 0.5f);
+
+            // ES START
+            // we cant set engine cvars by changing the file,
+            // and cant set client-only non-replicated cvars with configpresets
+            // so
+            _configManager.SetCVar("light.blur_factor", 0.0037f);
+            // ES END
         }
 
         public override void PostInit()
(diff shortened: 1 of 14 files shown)


mirrorcult:
need to check toast diff later just to see if theres anything weird. dont think so since i was checking it every time i committed but the diff size is larger than i expected

mirrorcult:
there will be a part 2 to this where i finish the lightswitch stuff and also tweak individual lights a bit more and whatevr

PR #342, opened 2025-12-03
change interaction outline colors for no reason

mirrorcult:
## About the PR

by 'no reason' i mean 'no practical reason but just because its kind of nice to have things be slightly different for vibes purposes'

in range
image

out of range
image


EmoGarbage404:
the tweak merchant collects her coin

PR #345, opened 2025-12-06
Redo emag

### LINKED ISSUE #221: EMAG should bolt doors open instead of a custom effect
## Description

The sparks and whatever are fine but EMAGing a door should be indistinguishable from normal tool manipulation or a station event.

We might want to keep generalizing unlocking for things like lockers, though? It might not be necessary.

EmoGarbage404:
## About the PR

Closes #221 

redoes emag with new code and less obvious effects.

summative changes:
- Access-readers no longer get their access wiped when emagged
- Emagging now sends out sparks

### DIFF
--- a/Content.Shared/_ES/Emag/ESEmagSystem.cs
+++ b/Content.Shared/_ES/Emag/ESEmagSystem.cs
@@ -1,0 +1,111 @@
+using Content.Shared._ES.Emag.Components;
+using Content.Shared._ES.Sparks;
+using Content.Shared.Access.Components;
+using Content.Shared.Administration.Logs;
+using Content.Shared.Charges.Systems;
+using Content.Shared.Database;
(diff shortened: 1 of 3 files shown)


Content.Shared/_ES/Emag/ESEmagSystem.cs
```diff
@@ -0,0 +1,111 @@
...
+        if (user.HasValue)
+        {
+            _popup.PopupPredicted(
+                Loc.GetString("emag-success", ("target", Identity.Entity(target, EntityManager))),
+                user.Value,
+                user,
+                PopupType.Medium);
+        }
+
+        _sparks.DoSparks(target, 3);
```
EmoGarbage404:
```suggestion
        _sparks.DoSparks(target);
```

PR #350, opened 2025-12-07
Add sabotage effect to tcomms servers

mirrorcult:
## About the PR

- saves radio messages sent to the channel in parrot memory (like polly)
- sends message on sabo. picks random message + picks a random name from a different random message, so the name and msg dont match
- doesnt distort, under the assumption that it might already have distortion added (both from the source msg and from when it sends). it doesnt in my video but
- parrot now saves nonrelayed radio messages also (thru activeradio) and also saves entity name

one technical issue is that the channels for listening and the channels the tcomms server are connected to are not technically synced, but i dont think theres any real way for it to get desynced anyway in a way that matters

https://github.com/user-attachments/assets/be4bbc9a-3d75-473a-82e5-473114d94111

### DIFF
--- a/Content.Server/Radio/EntitySystems/RadioSystem.cs
+++ b/Content.Server/Radio/EntitySystems/RadioSystem.cs
@@ -15,6 +15,11 @@ using Robust.Shared.Replays;
 using Robust.Shared.Utility;
 // ES START
 using Content.Server._ES.Radio;
+using Content.Server.Animals.Systems;
+using Content.Shared._ES.Degradation;
+using Content.Shared._ES.Masks.Traitor;
+using Content.Shared.Animals.Components;
+
 // ES END
 
 namespace Content.Server.Radio.EntitySystems;
@@ -31,6 +36,7 @@ public sealed class RadioSystem : EntitySystem
     [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly ChatSystem _chat = default!;
 // ES START
+    [Dependency] private ILocalizationManager _loc = default!;
     [Dependency] private readonly ESRadioSystem _esRadio = default!;
 // ES END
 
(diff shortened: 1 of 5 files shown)


Content.Server/Radio/EntitySystems/RadioSystem.cs
```diff
@@ -45,9 +50,46 @@ public override void Initialize()
...
+        // ES START
+        SubscribeLocalEvent<TelecomServerComponent, TransformSpeakerNameEvent>(OnTransformServerName);
+        SubscribeLocalEvent<TelecomServerComponent, ESUndergoDegradationEvent>(OnServerDegraded);
+        // ES END
+
         _exemptQuery = GetEntityQuery<TelecomExemptComponent>();
     }
 
+    // ES START
+    private void OnServerDegraded(Entity<TelecomServerComponent> ent, ref ESUndergoDegradationEvent args)
```
EmoGarbage404, line 59: needs to set args.handled = true for the degradation to complete properly (that's why the sparks aren't visible)

Content.Server/Radio/EntitySystems/RadioSystem.cs
```diff
@@ -15,6 +15,11 @@
 using Robust.Shared.Utility;
 // ES START
 using Content.Server._ES.Radio;
+using Content.Server.Animals.Systems;
+using Content.Shared._ES.Degradation;
+using Content.Shared._ES.Masks.Traitor;
+using Content.Shared.Animals.Components;
+
```
EmoGarbage404:
```suggestion
using Content.Shared.Animals.Components;
```
ominous space of doom

EmoGarbage404 (changes requested):
Commented on this in discord but if we were to make this a more continuous effect rather than a burst of messages, the way I'd like to see it is:
- Some kind of wire action for mending the broken server
- Maybe have like a modified sprite with random sparks when broken

Polly code already has all the fundamentals for this so I think that adding in the continuous behavior would be pretty quick and easy. Your call, though.

PR #352, opened 2025-12-07
remove 2nd hop beacon on toast + norot beacons

mirrorcult:
## About the PR

yep

image


Resources/Prototypes/Entities/Objects/Devices/station_beacon.yml
```diff
@@ -10,6 +10,7 @@
   - type: Sprite
     sprite: Objects/Devices/station_beacon.rsi
     drawdepth: BelowFloor
+    noRot: true
```
EmoGarbage404, line 14:
```suggestion
# ES START
    noRot: true
# ES END
```

PR #359, opened 2025-12-08
Fix comms console sabotage sound crashing

mirrorcult:
## About the PR

might have only been a crash on linux idk, i dont remember if the vfs always has case sensitivity


mirrorcult:
also i did kind of think these were linted but guess not

PR #364, opened 2025-12-08
Add degradation scheduler

EmoGarbage404:
## About the PR

random event scheduler that fires off degradation events. They're currently votable just for vibe purposes (though the low count in the pool means I had to restrict to like 2 options)

adds events for all planned sabo objectives + doors which i think generally covers traitor activity. Might need additional tuning but imo it's probably good for a first pass. This kind of plausible deniability is only kinda tangentially relevant to the first playtest.

### DIFF
--- a/Resources/Prototypes/_ES/GameRules/degradation_events.yml
+++ b/Resources/Prototypes/_ES/GameRules/degradation_events.yml
@@ -62,6 +62,8 @@
   - type: ESDegradationEvent
     component: TelecomServer
     degradeImmediately: true # Doesn't react to anything
+  - type: StationEvent
+    weight: 20
 
 - type: entity
   parent: ESBaseDegradationEvent
--- a/Resources/Prototypes/_ES/GameRules/degradation_events.yml
+++ b/Resources/Prototypes/_ES/GameRules/degradation_events.yml
@@ -1,0 +1,81 @@
+- type: entity
+  parent: BaseGameRule
+  id: ESBaseDegradationEvent
+  name: Unnamed Degradation Event
+  abstract: true
+  components:
+  - type: ESDegradationEvent
+  - type: StationEvent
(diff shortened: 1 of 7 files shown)


Princess-Cheeseballs:
It's missing the event where a 5'10" woman steps on you. This is a blocker.

mirrorcult:
> It's missing the event where a 5'10" woman steps on you. This is a blocker.

is this the tallest woman you could imagine before you got scared

Princess-Cheeseballs:
> is this the tallest woman you could imagine before you got scared

I'm bad at counting so yes. But in addition I also stole this joke from Emo because I'm evil.

Resources/Prototypes/_ES/GameRules/degradation_events.yml
```diff
@@ -0,0 +1,81 @@
...
+- type: entityTable
+  id: ESDegradationEvents
+  table:
+    all:
+    - id: ESDegradationEventAirlocks
+    - id: ESDegradationEventCommunicationsConsole
+    - id: ESDegradationEventCriminalRecords
+    - id: ESDegradationEventTelecoms
+    - id: ESDegradationEventCrewMonitoring
+    - id: ESDegradationEventIDCardComputer
```
mirrorcult, line 23: imo airlocks and telecoms should be weighted slightly higher than the others (since theres more of them), airlocks moreso

ArtisticRoomba:
> is this the tallest woman you could imagine before you got scared

The lion knows what he wants.

Resources/Prototypes/_ES/GameRules/degradation_events.yml
EmoGarbage404: done although ftr airlocks were already more common. they're just defined elsewhere because im piggybacking off station events for convenience.

PR #366, opened 2025-12-08
Add sabotage for ID card consoles

EmoGarbage404:
## About the PR

basic sabotage: when degraded, randomize the icon of 10-25% of IDs on station.

activates naturally whenever an inserted ID has its properties modified.

### DIFF
--- a/Content.Server/Access/Systems/IdCardConsoleSystem.cs
+++ b/Content.Server/Access/Systems/IdCardConsoleSystem.cs
@@ -152,7 +152,7 @@ public sealed class IdCardConsoleSystem : SharedIdCardConsoleSystem
         if (component.TargetIdSlot.Item is not { Valid: true } targetId || !PrivilegedIdIsAuthorized(uid, component, out var privilegedId))
             return;
 // ES START
-        _degradation.Degrade(uid, player);
+        _degradation.TryDegrade(uid, player);
 // ES END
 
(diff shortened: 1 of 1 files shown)


Content.Server/Access/Systems/IdCardConsoleSystem.cs
```diff
@@ -142,6 +151,9 @@ private void TryWriteToTargetId(EntityUid uid,
 
         if (component.TargetIdSlot.Item is not { Valid: true } targetId || !PrivilegedIdIsAuthorized(uid, component, out var privilegedId))
             return;
+// ES START
+        _degradation.Degrade(uid, player);
+// ES END
```
mirrorcult, line 156: huh? doesnt this immediately degrade it no matter if its sabotaged or not, shouldnt it just trigger the queued degradation or whatever?

mirrorcult (changes requested):

Content.Server/Access/Systems/IdCardConsoleSystem.cs
mirrorcult: or rather it should only trigger if theres `ESQueuedDegradation` i suppose

mirrorcult: ok looking at it it should just be trydegrade

PR #368, opened 2025-12-08
Add crew monitoring server sabotage

EmoGarbage404:
## About the PR

On degrade, randomize the suit sensors of everyone on the grid.

using the grid instead of the network stuff is mildly dud but also i think it's the simplest and the different is negligible and not relevant i think.


mirrorcult:
everyone on the grid seems like a lot but sure why not

PR #374, opened 2025-12-10
Restore job outfits

### LINKED ISSUE #284: Passenger should be renamed to Assistant
"Passenger" is a sauceless name. most people call them assistants because the name is cooler. Also they actually will be assisting people now in theory with the advent of the station objective.

Do not rename the internal prototypes if you do this. Only rename the player-facing strings.

EmoGarbage404:
## About the PR

Fixes missing job clothing on default outfits

Closes #284


Resources/Prototypes/Entities/Markers/Spawners/jobs.yml
```diff
@@ -91,7 +91,9 @@
 - type: entity
   id: SpawnPointPassenger
   parent: SpawnPointJobBase
+# ES START
   name: passenger
+# ES END
```
mirrorcult, line 96: no actual change here?

PR #381, opened 2025-12-12
Make the antimatter not instantly evaporate mobs

### LINKED ISSUE #548: Antimatter instantly kills people
### What version did the issue occur in?

_No response_

### Description

It instantly eats your body. This also applies to Phantom ghosts and will instantly eat the ghost, sending them to the lobby.

### Reproduction

1. walk into antimatter
2. wake up in lobby

### Additional Context

_No response_

PotatOverlord:
## About the PR

Closes #548 

if something has a threshold it will take 5 cellular damage until it dies and is then consumed
## Why / Balance

it says to do this in the doc and it's not like that

## Technical details

uses the IsDead api from MobThresholdSystem in EventHorizonSystem in an event handler to only consume it when its not dead

## Media

todo

## Requirements

- [ X ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ ] I have added media to this PR or it does not require an in-game showcase.

Idk how to do c# properly this might be cooked. It worked when I was testing it though
Also needs some sort of way of alerting the player to the fact that they are being killed probably


FlippenPage:
the added bool wont work because its not used anywhere in the method.
you can access the eventhorizon component through event args to grab it here.

PotatOverlord:
I'm lowk kinda stupid guh

EmoGarbage404 (changes requested):
From a cursory glance, does this even work correctly when a mob dies in the antimatter? I feel like it's not going to attempt to consume them until they move and retrigger collision events.

Content.Shared/Singularity/Components/EventHorizonComponent.cs
```diff
@@ -82,4 +82,13 @@ public sealed partial class EventHorizonComponent : Component
     public TimeSpan NextConsumeWaveTime;
 
     #endregion Update Timing
+
+    //ES Start
+    /// <summary>
+    /// If this event horizon can consume alive mobs.
+    /// </summary>
+    [DataField]
+    public bool consumeAliveMobs = true;
```
EmoGarbage404, line 91: should be `ConsumeAliveMobs` (capitalize first letter)

Content.Server/Singularity/EntitySystems/EventHorizonSystem.cs
```diff
@@ -402,6 +412,17 @@ public static void PreventBreach<TComp>(EntityUid uid, TComp comp, ref EventHori
...
     }
 
+    // ES Start
+    public void PreventConsumeAlive(Entity<MobStateComponent> ent, ref EventHorizonAttemptConsumeEntityEvent args)
+    {
+        if (args.Cancelled)
+            return;
+        if (!MobStateSystem.IsDead(ent.Owner, ent.Comp) && args.EventHorizon.consumeAliveMobs)
+            PreventConsume(ent.Owner, ent.Comp, ref args);
+    }
```
EmoGarbage404: i think this is confusingly written (hence the error) when it should look more like
```suggestion
        if (args.Cancelled || args.EventHorizon.consumeAliveMobs)
            return;
        if (MobStateSystem.IsDead(ent.Owner, ent.Comp))
            return;
        PreventConsume(ent.Owner, ent.Comp, ref args);
    }
```

Content.Shared/Singularity/Components/EventHorizonComponent.cs
PotatOverlord: The old way of declaring datafield names was throwing me off, thanks

EmoGarbage404:
i think my solution is like 50% a hack but its a necessary one so i'm just gonna merge and if it's shit you can hang me later for it.

PR #382, opened 2025-12-12
Objective Rewrite

### LINKED ISSUE #328: Rewrite objectives
Christmas project for me but im sick of fucking looking at objective code

checklist:
- put that shit in shared. i'm begging.
- Cache objective information instead of having to query it manually
  - Add some "refresh objectives of X type on X mind" call
  - The recalculated objective progress is cached on a comp and then networked
  - we can ACTUALLY UPDATE THE UI WHEN THE OBJECTIVE CHANGES!!!11!!!
- Make selection 10x more sane
  - might need a custom condition system or maybe entity table conditions can suffice. i dont care which.
- Make objectives support richtext and for the love of god make the ui look better.
- EDIT: clamp progress on [0, 1]. genuinely can't believe it doesn't do this already.

### LINKED ISSUE #361: Kill objectives dont work
Image

that, or they just don't display as completed?

EmoGarbage404:
## About the PR

Closes #328 
Closes #361

big rewrite of objectives. not done but throwing it here to bigdick

todo:
- [x] kill objective
- [x] arms dealer objective
- [x] test refactored objectives (would be appreciated if someone helped me out on it)


Content.Shared/_ES/Objectives/ESSharedObjectiveSystem.cs
```diff
@@ -0,0 +1,280 @@
...
+    [PublicAPI]
+    public List<Entity<ESObjectiveComponent>> GetObjectives(Entity<ESObjectiveHolderComponent?> ent)
+    {
+        return GetObjectives<ESObjectiveComponent>(ent);
+    }
+
+    /// <summary>
+    /// Returns all objectives on an entity which have a given component
+    /// </summary>
+    public List<Entity<T>> GetObjectives<T>(Entity<ESObjectiveHolderComponent?> ent) where T : Component
```
moonheart08, line 160: These two seem a lot like queries and should probably return a concrete implementer of IEnumerable, without allocating, instead. (Times have changed since entity queries were written the way they are so manual movenext isn't useful perf wise anymore, just make sure C# actually knows the concrete type so don't return IEnumerable.)

Just so in the future if someone uses these like queries, they don't spam a bunch of allocs in a hot loop somewhere.

mirrorcult:
ok well gh doesn't like me reviewing on my phone I guess so I don't think the individual reviews went through. what I remember
- i don't like the ergonomics in the event handlers for checking if an objective is valid to assign (having to remember to check and being able to accidentally override), imo it should work more like cancelable events where args.Valid is private set and handlers just call something like args.Invalidate() on their fail condition
- it feels weird to still kind of have that objective progress indirection, like the objectives rn sub to some event or whatever then call refresh progress then have a separate handler for refresh progress that sets a field on the event to get used for the actual stored progress when I feel like it would make more sense to just have them set progress directly or something. having the event implies there would be multiple sources of progress on the same objective entity which just seems nonsensical?

Issue #396, opened 2025-12-14
Bridge & Captain's Quarters on the arrivals shuttle is all-access


Flareguy:
oops

PR #435, opened 2025-12-14
Various arrivals fixes

### LINKED ISSUE #422: Criminal records console on the arrivals shuttle should be removed / replaced
It's a very easy objective target, since nobody goes there.

Flareguy:
## About the PR
Resolves #422 and #396.

Also changes the bar to be slightly less weird and less cheesable for the guzzler, replaces the clothesmate with a solar's hot drinks, and makes the captain's quarters airlock not glass since that sightline didn't make much sense at all.

## Media
Content Client_jU5TARi91O

## Requirements

- [X] I have added media to this PR or it does not require an in-game showcase.

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Maps/_ES/Shuttles/arrivals.yml
+++ b/Resources/Maps/_ES/Shuttles/arrivals.yml
@@ -4,8 +4,8 @@ meta:
   engineVersion: 268.1.0
   forkId: ""
   forkVersion: ""
-  time: 12/14/2025 07:48:02
-  entityCount: 727
+  time: 12/14/2025 16:57:16
+  entityCount: 725
 maps: []
 grids:
 - 1
@@ -223,7 +223,7 @@ entities:
       parent: 1
 - proto: AirlockCaptainLocked
   entities:
-  - uid: 229
+  - uid: 188
     components:
     - type: Transform
       pos: 35.5,-4.5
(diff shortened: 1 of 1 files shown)


mirrorcult:
the rechargers should probably also be removed for the same reason as the records computer, kind of way too easy to get to and not actually useful while you're on arrivals (although i want arrivals to get locked off anyway)

Flareguy:
> the rechargers should probably also be removed for the same reason as the records computer, kind of way too easy to get to and not actually useful while you're on arrivals (although i want arrivals to get locked off anyway)

i mean, they *are* pretty useful if you need to actually detain someone. i think i'm fine with them existing, since there are a lot of other fairly easy ways to get rechargers & if arrivals gets locked off then it'll probably be about as difficult as abusing a secpost or whatever.

EmoGarbage404:
I would rather the arrivals shuttle doesnt have utility stuff like rechargers, tasers, batons, etc. because secoffs already spawn with this stuff on their belt and it makes arrivals a giant target due to low traffic.

Flareguy:
everything has been removed except the handcuffs, since i figured that those were mostly inoffensive and also pretty useful for non-security to have on the shuttle. they're definitely a valuable target, but one i think i'm fine with having since most people would probably just want to craft zipties anyways or something. i see it as about equivalent to the drinkable surplus that exists.

PR #453, opened 2025-12-15
Revert "Draw viewcone in WorldSpaceBelowFOV"

mirrorcult:
Reverts EphemeralSpace/ephemeral-space#442

i have no fucking idea why but this makes lighting significantly darker for some reason??? ???????????????????


mirrorcult:
i genuinely thought i was going insane

moonheart08:
Worth triaging with renderdoc.

mirrorcult:
image
image

comparison (top is incorrect bottom is correct)

EmoGarbage404:
oh that's what that was. i noticed it when testing and was like "wow this is some insane variation pass" and then promptly forgot because it was late.

PR #454, opened 2025-12-15
Better lightglow override support on bulbs

mirrorcult:
## About the PR

removes my old glowalpha thing
allows fully overriding light glow layer color on bulbs
overrides the color for our basic lights so they stand out more compared to their actual glow color. i like how it looks better

image


EmoGarbage404:
tweak merchant

PR #480, opened 2025-12-18
Empath mask

EmoGarbage404:
## About the PR

From doc https://ephemeralspace.github.io/docs/design/masks/crew/empath.html

empaths can use an ability every 5 minutes to read the aura off of one person. the aura corresponds to their objectives loosely.
For any objectives that involve _killing_ (veteran, phantom, traitors, etc.), their aura reads as a "killing intent." Everyone else reads as a neutral aura. Strange aura can be added trivially once amnesiac or copier exists.


Content.Shared/_ES/Masks/Aura/ESSenseAuraSystem.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// <inheritdoc/>
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<ESSenseAuraEvent>(OnSenseAuraEvent);
+    }
+
+    private void OnSenseAuraEvent(ESSenseAuraEvent args)
+    {
+        // This can't be predicted but there's no harm in having it in shared.
+        if (!_mind.TryGetMind(args.Target, out var mind, out _))
```
moonheart08: This whole thing feels.. overly abstract to me? I get the intent but there should probably be an override for it on individual masks (not on the aura prototype), which would also solve the "no objectives" problem.

Additionally, if being generalized this way, the senser should probably have a list of auras they recognize so you can have auras for use only by other aura-based masks (empath probably doesnt need to recognize a "fruity" aura, while the Fruit Destroyer probably does.)

moonheart08 (changes requested):

PR #481, opened 2025-12-18
17 12 2025 masks reorganization

moonheart08:
Git really doesn't like the diff but I promise I jut alphabetized the masks.


Resources/Prototypes/_ES/Masks/traitors.yml
```diff
@@ -0,0 +1,42 @@
...
+  abstract: true
+  components:
+  - type: ESTroupeFactionIcon
+    icon: ESTraitor
+    troupe: ESTraitor
+    examineString: es-mask-troupe-traitor-examine
+
+- type: esMask
+  parent: ESBaseTraitor
+  id: ESMarauder
```
EmoGarbage404, line 16: should be at the end, i would think

Resources/Prototypes/_ES/Masks/crew.yml
```diff
@@ -73,6 +42,20 @@
...
+  weight: 1
+  objectives:
+    id: ESObjectiveEatSummonedFruit
+  components:
+  - type: ActionGrant
+    actions:
+    - ESActionMaskSpawnFruit
+
 - type: esMask
   id: ESGuzzler
```
EmoGarbage404, line 60: should be after glutton

EmoGarbage404 (changes requested):
>alphabetize masks
>didn't alphabetize them

EDIT: also while you're here it would be nice to potentially break up the objectives into individual files corresponding to each mask, rather than all of them being lumped in to the "crew.yml" objectives file

EmoGarbage404:
shot dead in the streets by 10 gangsters in fedoras

PR #504, opened 2025-12-19
Maruader & Infiltrator Cache Variation

Ermucat:
## About the PR

Adds two new possible pools to infiltrator and maruader caches
## Why / Balance

Variation is good and increases replay ability
## Technical details

Edits the pools of ESTraitorCacheInfiltrator and ESTraitorCacheMarauder
## Media

image

## Requirements

- [X] I have added media to this PR or it does not require an in-game showcase.

## Breaking changes

**Changelog**

:cl:
- tweak: Maurader and Infiltrator can now have varying loot in their caches


mirrorcult:
the marauder variation seems fine but i'd prefer the infiltrator to always have an emag i think + scram doesnt fit much i think

EmoGarbage404 (changes requested):
thoughts:
- agree with mirror that infiltrator should probably always have the EMAG as their main equipment piece
- The C4 is honestly a better fit for the infiltrator than the smoke grenade. I think it's more useful in general
- For all the grenade changes, i think it would be better if you just did a single group selector with all the grenades and then made it roll 2 times. That would give you a sorta random selection of a few different grenades which seems more interesting. 

In general I would prefer if the randomization was less "Kit A or Kit B" and more like "you get one melee out of a few and 3 grenades out of a selection." I think any substantially different kits should eventually just be unique masks.

Sadie-silly:
i dont like esword being in there at all, considering it's like a prime "This person is a bad guy" item

EmoGarbage404:
> i dont like esword being in there at all, considering it's like a prime "This person is a bad guy" item

that is true. You do kinda completely lose any deniability (although there's still overlap with mercenary and other innate valid items so maybe it's fine?)

Sadie-silly:
> > i dont like esword being in there at all, considering it's like a prime "This person is a bad guy" item
> 
> that is true. You do kinda completely lose any deniability (although there's still overlap with mercenary and other innate valid items so maybe it's fine?)

eehh,, maybe, though i'd still rather have it be some sort of weapon that is available to most anyone if they're dedicated enough. Even if mercenary can get it, it still instantly reduces anyone holding it to traitor or mercenary

Flareguy (changes requested):
i appreciate the idea behind this PR but i think the items used feel rather scattershot and wouldn't align very well with our minimalist approach:

- i deliberately avoided including the energy sword in the marauder's initial kit since it has bad plausible deniability and is probably too powerful.
- stinger grenades are very weird feeling and security is definitely not going to have them, so the marauder should probably not get them
- like other people have said, the alternative infiltrator kit is probably too different
- in general i think i would rather individual elements get swapped out with equivalent items instead of having different sub-kits like it is set up now

PR #505, opened 2025-12-19
Add electrical fire station event

EmoGarbage404:
## About the PR

adds a random station event that starts a fire somewhere on the station. Has stagehand integration for voting where the fire will start.

image


mirrorcult (approved):

mirrorcult:
might end up being too strong but we'll see

PR #508, opened 2025-12-19
Change fires to use TileEmission

EmoGarbage404:
## About the PR

dear god the lighting budget

OLD:
image

NEW:
image

this is a bit less intense but maybe it's still good? I could go in and add some manual code that adjusts the range and color on TileEmission if that's wanted. I want mirrors opinion on it tho, since it's really her ballpack and the general effect of increasing energy can't be replicated with this method.

### DIFF
--- a/Resources/Prototypes/_ES/Entities/Effects/tile_fires.yml
+++ b/Resources/Prototypes/_ES/Entities/Effects/tile_fires.yml
@@ -28,14 +28,6 @@
     layers:
     - map: [ "base" ]
       state: red_1
-  - type: PointLight
-    netsync: false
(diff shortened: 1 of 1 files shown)


EmoGarbage404:
image

alternative visuals with an unshaded layer + a brighter color for the lights

EDIT: this looks way better so im pushing this to the PR

mirrorcult:
> image
> 
> alternative visuals with an unshaded layer + a brighter color for the lights
> 
> EDIT: this looks way better so im pushing this to the PR

this is satisfactory imo

mirrorcult (approved):

PR #516, opened 2025-12-20
Sacrifice mask

mirrorcult:
## About the PR

https://ephemeralspace.github.io/docs/design/masks/crew/sacrifice.html

so, this one is a little interesting, because the logic for doing the healing, and the objective logic for counting the heals are actually tied to the same thing. meaning having the objective is what controls having the healing-on-death. but also like. this just seems extremely reasonable and is probably a good way to have a lot of stuff work (thinking of polymath here)?

https://github.com/user-attachments/assets/0789a128-5af4-4094-b255-8938d15e4509


Content.Server/_ES/Masks/Objectives/Components/ESSacrificeHealObjectiveComponent.cs
```diff
@@ -0,0 +1,34 @@
...
+/// </summary>
+/// <seealso cref="ESSacrificeHealObjectiveSystem"/>
+[RegisterComponent]
+public sealed partial class ESSacrificeHealObjectiveComponent : Component
+{
+    /// <summary>
+    ///     Mob state that will trigger the effect & objective completion.
+    /// </summary>
+    [DataField]
+    public MobState TriggerMobState = MobState.Dead;
```
moonheart08: Feels overly generic, in that I'm not sure how this would be useful set to any other value.

Content.Server/_ES/Masks/Objectives/ESSacrificeHealObjectiveSystem.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    private void OnMobStateChanged(Entity<ESSacrificeHealObjectiveComponent> ent, ref ESMobStateChanged args)
+    {
+        if (args.NewMobState != ent.Comp.TriggerMobState)
+            return;
+
+        var coords = Transform(args.Mob.Owner).Coordinates;
+
+        var count = 0;
+        var msg = Loc.GetString("es-sacrifice-popup-heroic-sacrifice",
```
moonheart08: Which localization to use should be provided by the component as a datafield.

moonheart08 (changes requested):

Content.Server/_ES/Masks/Objectives/Components/ESSacrificeHealObjectiveComponent.cs
mirrorcult: fair enough

moonheart08 (approved):

PR #589, opened 2025-12-21
Fix disappearing objectives

### LINKED ISSUE #542: Objectives broke after disconnecting and reconnecting
wasn't able to see them after hopping through. PVS often has weird fuckery so im not surprised.

EmoGarbage404:
## About the PR

Closes #542 

session overrides dont persist upon client disconnect so it would just stop sending you the data for objectives. Fixed it with a pretty basic relay and some extra events.


EmoGarbage404:
after merging this the beauty of the universe revealed itself to me and i figured out a better way to do this behavior

PR #599, opened 2025-12-22
Playtest 2 mask rebalancing

### LINKED ISSUE #552: Glutton Objectives are too hard
You need to eat too many unique items to complete it

### LINKED ISSUE #554: Daredevil objectives are too hard
total damage, group damage, damage sources. all of it is way too fucking high.

### LINKED ISSUE #577: Vandals are too common
stopgap for masquerades, make it like 0.5

mirrorcult:
## About the PR

fixes #577 
fixes #552 
fixes #554 

reduced the targets for a lot of the freak masks quite substantially, would prefer these to be slightly lower than they ought to be rather than way higher

massively reduces weight for vandal, and slightly reduces weight for martyr (to the same weight as phantom) for now

### DIFF
--- a/Resources/Prototypes/_ES/Masks/masks.yml
+++ b/Resources/Prototypes/_ES/Masks/masks.yml
@@ -159,6 +159,7 @@
   troupe: ESCrew
   description: es-mask-vandal-desc
   color: tomato
+  weight: 0.2
   mindComponents:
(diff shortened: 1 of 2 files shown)


EmoGarbage404:
aligns with what we chatted about

PR #605, opened 2025-12-22
More effective pulling (hack shit) + dont occlude pulled ents

mirrorcult:
## About the PR

- store an entities fixtures density and then set their density artificially low on pull
- restore it to the old density on stop pull
- on client, add comp to occludables that get pulled to never occlude them (looks better this way and i dont want to do the 'you face stuff you pull' thing because backwalking means it would slow you down and i dont like that)

https://github.com/user-attachments/assets/f586c1fb-7b4a-4f30-b65b-449002888cbb


mirrorcult:
also reducing pull range more because it sickens me

mirrorcult:
and i have deep hatred in my heart

PR #612, opened 2025-12-23
Distributed cargo spawner

### LINKED ISSUE #356: Add a distributed spawner for cargo initial spawns
currently, cargo uses a slew of random crate spawners to fill out their initial set of cargo crates in the warehouse. This is Fine but using distributed spawners we can do something a lot better.

create a distributed spawner that just has a consistent amount of crates and other goodies in it. you can add mostly whatever since there really isnt good guidelines right now. I just want the spawner for maintenance reasons.

also make sure the range on it is big enough to cover all of cargo.

CrystalineSpider:
## About the PR
Resolves #356

No mapping changes yet, just the spawner and a selection of crates for it.

Later we can make it weighted, but I wanted to at least get the core pushed.
## Media
Screenshot 2025-12-22 174019
Screenshot 2025-12-22 174036

## Requirements

- [x] I have added media to this PR or it does not require an in-game showcase.


Resources/Prototypes/_ES/Catalog/Crates/crates.yml
```diff
@@ -0,0 +1,279 @@
...
+- type: entity
+  parent: CrateEngineeringSecure
+  id: ESAdvancedToolsCrate
+  name: advanced tool crate
+  description: With these you might almost feel ready for a promotion!
+  components:
+  - type: EntityTableContainerFill
+    containers:
+      entity_storage: !type:AllSelector
+        children:
```
EmoGarbage404:
```suggestion
      entity_storage:
        all:
```
You can do this to shorten it a bit and make it easier to read (this can be done across most of this file)

Resources/Prototypes/_ES/Entities/Markers/Spawners/Distributed/cargo.yml
```diff
@@ -0,0 +1,50 @@
+- type: entity
+  id: ESSpawnerRandomDistributedCargo
+  parent: ESMarkerBase
+  name: Cargo Crate Spawner
+  components:
+  - type: Sprite
+    layers:
+    - state: crate
+    - state: random
```
EmoGarbage404:
```suggestion
    - state: rand-dist
```

as a distributed spawner, should use the `rand-dist` state instead of `random`

Resources/Prototypes/_ES/Entities/Markers/Spawners/Distributed/cargo.yml
```diff
@@ -0,0 +1,50 @@
+- type: entity
+  id: ESSpawnerRandomDistributedCargo
+  parent: ESMarkerBase
+  name: Cargo Crate Spawner
+  components:
+  - type: Sprite
+    layers:
+    - state: crate
```
EmoGarbage404, line 68: the crate icon is fairly generic so you should add a new sprite to the spawner rsi that says "CARGO" or something so that they're easier to distinguish. There's a ton of crate spawners and having it denoted as specifically for cargo would improve readability when mapping.

EmoGarbage404 (changes requested):
didnt really look at the crate fills beyond a glance. we could use more coverage but that's easy to add onto once we have this part generally ironed out.

Resources/Prototypes/_ES/Entities/Markers/Spawners/Distributed/cargo.yml
CrystalineSpider: I wasn't sure how the sprites worked so left the crate as mostly a default, I'll work on a proper label though since I do agree

CrystalineSpider:
> didnt really look at the crate fills beyond a glance. we could use more coverage but that's easy to add onto once we have this part generally ironed out.

What do you mean?

EmoGarbage404:
> What do you mean?

there's a lot more potential crates we could add, but I think this is probably fine for a first selection since we're really just hashing out the basics for now

CrystalineSpider:
> there's a lot more potential crates we could add, but I think this is probably fine for a first selection since we're really just hashing out the basics for now

That was mostly my thought. I haven't played much cargo and don't really know what people need/want but I figured we could always do a second pass once the framework is there.

CrystalineSpider:
added weights to the items so the new sec crates wouldn't make up the entire cargo bay and resolved your requested changes. Also added a few crates, mostly various security crates and a few other lower weight ones.

Resources/Prototypes/_ES/Entities/Markers/Spawners/Distributed/cargo.yml
```diff
@@ -0,0 +1,74 @@
...
+    - state: rand-dist
+    - state: cargo
+    - state: es
+    - state: c100
+  - type: EntityTableSpawner
+    table: !type:NestedSelector
+      tableId: ESSpawnerCargo
```
EmoGarbage404: this isn't using a distributed spawner anymore, which needs to be fixed. it should look something like

```yml
  - type: ESDistributedSpawner
    id: ESSpawnerCargo
    tables:
    - tableId: ESSpawnerCargo
    - tableId: ESSpawnerCargo
    - tableId: ESSpawnerCargo
    - tableId: ESSpawnerCargo
    - tableId: ESSpawnerCargo
 ```

realistically in the future we should probably weight this so it's less "20 fully random crate spawners" and more like "2 of this group, 5 of this group, 10 of this group, etc."

EmoGarbage404 (changes requested):

Resources/Prototypes/_ES/Entities/Markers/Spawners/Distributed/cargo.yml
CrystalineSpider: If I may ask, whats the purpose of having the tableId replicated?

Also I agree doing weights with certain amounts of various crate sets would be ideal, and while I might look into that later it's functional enough in my opinion as is. (Most of this was based off of me seeing examples of functions in other yaml files and I just didn't find a good example of that kind of randomization, if I find one though I will likely return to revise this again)

CrystalineSpider: Actually I think I figured out the way to weigh this to have less random options so I'll get back to this.

CrystalineSpider:
I hope this works, changes TLDR because I feel formal, but spawner now spawns a set amount of each "department" of crate, such as four material-related crates or two medical-related crates.

PR #622, opened 2025-12-24
Minor viewcone tweaks

mirrorcult:
## About the PR

less feathering
slightly less surround radius
slightly darker

image

the grain isnt act ually this crazy its just an artifact of the fact that its always going to look grainy when you take a static screenshot vs video


mirrorcult:
image
old

PR #627, opened 2025-12-24
Screenshake

mirrorcult:
## About the PR

drafting for my own sake mostly

the base work is done i mostly just have to go thru and tune all the numbers so it feels good and add it to a bunch of things that need it


WinkRussian:
finally turning my space game into hotline miami

Content.Shared/_ES/Camera/ESScreenshakeSystem.cs
```diff
@@ -0,0 +1,196 @@
...
+            }
+        }
+    }
+
+    private void OnGetEyeOffset(Entity<ESScreenshakeComponent> ent, ref GetEyeOffsetEvent args)
+    {
+        if (!TryComp<EyeComponent>(ent, out var eye))
+            return;
+
+        var seed = SharedRandomExtensions.HashCodeCombine((int)_timing.CurTick.Value, GetNetEntity(ent).Id);
```
moonheart08: This should probably be a SmallRandom instead. Though I'm having trouble figuring out what you're doing here with a FastNoiseLite exactly?

Can't really figure out what properties of the noise you're actually relying on, given the fact you change the seed so often that it should have a perfectly random value.

You can't netsync this (and thus shouldn't use the janky "networked rng" trick) because it needs to be continuous over frames, not ticks, unless you plan to make this update tick-wise instead and then interpolate over multiple frames. ATM this will effectively run at Update()'s rate instead of per frame because the rng will be the same over multiple frames.

Content.Shared/_ES/Camera/ESScreenshakeSystem.cs
```diff
@@ -0,0 +1,196 @@
...
+        args.Offset += accumulatedOffset;
+    }
+
+    private void OnGetEyeRotation(Entity<ESScreenshakeComponent> ent, ref ESGetEyeRotationEvent args)
+    {
+        if (!TryComp<EyeComponent>(ent, out var eye))
+            return;
+
+        // +10 to avoid reusing same values as other shit
+        var seed = SharedRandomExtensions.HashCodeCombine((int)_timing.CurTick.Value, GetNetEntity(ent).Id) + 10;
```
moonheart08: ditto.

Content.Shared/_ES/Camera/ESScreenshakeSystem.cs
mirrorcult: noise is for a not-discontinuous shake + more tweakability

i dont actually think this even needs to be networked anymore, i think it was in an older version but not now ? so i can probably just remove the networked rng stuff entirely

mirrorcult:
#yup #undrafted

neomoth:
Too lazy to PR this here myself but if you guys still have the issue with the camera being rotated as things load in for a split second both at roundstart and at connection, add this to `Content.Client.Eye.EyeLerpingSystem.AddEye`. Seems to fix it for us.
```cs
public void AddEye(EntityUid uid, EyeComponent? component = null, bool automatic = false)
{
    if (!Resolve(uid, ref component))
        return;

    var lerpInfo = EnsureComp<LerpingEyeComponent>(uid);
    lerpInfo.TargetRotation = GetRotation(uid);
    lerpInfo.LastRotation = lerpInfo.TargetRotation;
    lerpInfo.ManuallyAdded |= !automatic;

    lerpInfo.TargetZoom = component.Zoom;
    lerpInfo.LastZoom = lerpInfo.TargetZoom;

    if (component.Eye != null)
    {
        _eye.SetRotation(uid, lerpInfo.TargetRotation, component);
        _eye.SetZoom(uid, lerpInfo.TargetZoom, component);
        //ADD HERE ADD HERE ADD HERE ADD HERE ADD HERE
        if (TryComp<ContentEyeComponent>(uid, out var contentEye))
            contentEye.BaseRotation = lerpInfo.TargetRotation;
        //ADD HERE ADD HERE ADD HERE ADD HERE ADD HERE
    }
}
```

mirrorcult:
don't think I've ever noticed that happening but I'll keep it in mind if I do, thank you

PR #637, opened 2025-12-25
no shuttle world

### LINKED ISSUE #607: Communications console should be unable to call the shuttle
No going back, little man.

mirrorcult:
## About the PR

closes #607 

disables calling/recalling from the comms console
disables shuttle call ability completely 

i would have removed the button, but i like it being there just greyed out. for themes and writing reasons

image


EmoGarbage404:
test fail isn't erroneous since apparently shuttle calling has a test to ensure that it always works.

PR #647, opened 2025-12-27
Red Carpet + Traitors (new high weight mode)

moonheart08:
## About the PR
Adds two new masquerades, one to serve as the new "balanced traitors round" and another to devalue the VIP economy.
## Why / Balance
Random is inherently not a balanced masquerade, `Traitors` actively tries to be and is 100% worth bikeshedding a bit. Red Carpet is a variant on Traitors with an outsized excess of VIPs.

Will conflict with #721

### DIFF
--- a/Resources/Prototypes/_ES/Masquerades/traitors.yml
+++ b/Resources/Prototypes/_ES/Masquerades/traitors.yml
@@ -8,10 +8,10 @@
   masquerade: !type:MasqueradeRoleSet
     defaultMask: "ESCrewmember" # God help you.
     roles:
-      6: ["ESMarauder(2)", "#Oracles", "#Softsafes", "#Gunners"] # 1 generic crew, 2 traitors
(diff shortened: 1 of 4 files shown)


Resources/Prototypes/_ES/Masquerades/traitors.yml
```diff
@@ -0,0 +1,23 @@
...
+  name: Traitors
+  description: standard, balanced cast of crew and traitors
+  weight: 10
+  minPlayers: 6
+  gameRules: []
+  masquerade: !type:MasqueradeRoleSet
+    defaultMask: "ESCrewmember" # God help you.
+    roles:
+      6: ["ESMarauder(2)", "#Oracles", "#Softsafes", "#Gunners"] # 1 generic crew, 2 traitors
+      8: ["#Softsafes"] # 2 generic crew, 2 traitors.
```
mirrorcult, line 12: nit but think id prefer marauder+infiltrator rather than double marauders for lowpop, seems too snowbally

moonheart08: Perhaps a Recruit?

PR #648, opened 2025-12-27
Feat/telesci

EmoGarbage404:
## About the PR

Basic bitch telescience.

bare minimum stuff i want for the playtest.


mirrorcult:
the telescy is real

PR #653, opened 2025-12-27
Disable greencomms

EmoGarbage404:
## About the PR

experimental test that didnt really shake out. The additional limitation on communication was sorta nice but it also just made it hard for people to communicate and was a general pain.


mirrorcult:
this is really "enable greencomms"

EmoGarbage404:
> this is really "enable greencomms"

i dont know what the term greencomms really means so im just saying whatever

PR #690, opened 2025-12-28
Add two trash bags and extra bucket to janitor closet

### LINKED ISSUE #665: Janitor has no trash bag

K-Dynamic:
## About the PR

Adds two trash bags and one extra bucket to Toast's janitor closet

## Why / Balance

Fixes https://github.com/EphemeralSpace/ephemeral-space/issues/665.

Second bucket is needed since there's two janitors.

## Media

image

## Requirements

- [x] I have added media to this PR or it does not require an in-game showcase.

**Changelog**

:cl:
- tweak: Added two trash bags and an extra bucket to the janitor closet.

### DIFF
--- a/Resources/Maps/_ES/toast.yml
+++ b/Resources/Maps/_ES/toast.yml
@@ -4,8 +4,8 @@ meta:
   engineVersion: 268.1.0
   forkId: ""
   forkVersion: ""
-  time: 12/27/2025 20:17:19
-  entityCount: 14586
+  time: 12/28/2025 03:50:45
+  entityCount: 14589
 maps:
 - 3
 grids:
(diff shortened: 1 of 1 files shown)


CrystalineSpider:
There is a bucket in the room below the closet (or at least was at one point), so if this is pushed through I will go remove that one as it definitely is a better position. As for trash bags, yeah definitely a good move.

K-Dynamic:
It might be better for the secure trash cart to have trashbags inside but this will do for now

EmoGarbage404:
isn't there a custodial closet prototype or something that spawns with some of this stuff? I'd rather we either use a closet or make some spawners or something over throwing down extra stuff on a rack since i think it's not very scalable.

Issue #716, opened 2025-12-28
End of round screen should show the masquerade


EmoGarbage404:
people were confused about what showdown was since i dont think it was telegraphed very well.

PR #719, opened 2025-12-28
Masquerade fixes

### LINKED ISSUE #692: Freakshow doesnt work
crashes on startup

@moonheart08

moonheart08:
- Fixed an issue where all pop levels besides max pop would get bogus mask assignments.
- Fixed an issue where excess masks on a player count would cause a hang.
- Removed legacy random logic in favor of a new system.
- Expand mask sets to support more than just a weighted dict.
- ApplyMask can now start troupe rules on its own.

Fixes #692


moonheart08:
in an effort to avoid another massive PR, this is now reviewable. Will handle #716 in another PR.

moonheart08:
Rebased and ready 2 go

Content.Server/_ES/Masks/ESMaskSystem.cs
```diff
@@ -236,36 +170,19 @@ public bool IsPlayerValid(ESTroupePrototype troupe, ICommonSession player)
...
-            if (maskProto.Troupe != troupe)
-                continue;
-
-            // TODO: check the mask has valid objectives.
-            // Don't assign masks if their objectives can't be done.
-
-            weights.Add(maskProto, maskProto.Weight);
+            if (!TryGetTroupeEntityForMask(mask, out troupe))
+            {
+                _gameTicker.StartGameRule(PrototypeManager.Index(mask.Troupe).ID, out var troupeEnt);
```
EmoGarbage404: the troupe should really have a gamerule entity on the prototype rather than assuming one exists with the same ID as the troupe.

EmoGarbage404 (approved):
minor nit

Content.Server/_ES/Masks/ESMaskSystem.cs
moonheart08: ah hell. this was a bug/case of not thinking. oops.

PR #720, opened 2025-12-28
2025-12-28 upmerge

mirrorcult:
conflicts were quite sane this time around

relevant notes for contributors going forward
- .net 10 is required now
- the solution file was moved to `.slnx`, you will need to reopen the correct file with your IDE after updating

upmerge issues that should be noted
- sleepypen stuff got changed bc of the hypospray->injector merge. idk if it works the same way as it used to, check 6a280f85311c193676463a672187dfcedfb97745 and whatnot. it has awful UX now but maybe it was always like that and this api is kinda not great so its hard to figure out how to change
- integrationtests did not like importing jetbrains annotations for some reason so i just removed the annotations in the new gametest stuff d5d2f368152b27c09338d2300087b889a7f9c2b7 @moonheart08
- masquerade prototype stuff was datadefs but not partial. that wasnt crashing before now. IDK. who cares i guess. also i had to change all our prototype id/parent/abstract fields to be `get; private set;` since thats required now i guess
- deletes old es med stuff in `Prototypes/_ESP` since merge issues and we dont use them anyway


mirrorcult:
and do not squash obv

moonheart08:
No qualms with the change to gametest, I'll look at it once upmerge is.. merged.

PR #733, opened 2025-12-29
Disable this test properly

EmoGarbage404:
## About the PR

it has some insane shit so even if the assert is disabled it fails anyways.


ArtisticRoomba:
real

moonheart08 (commented):
Instead of bodging the test further, maybe just mark it `[Explicit]` or `[Retry(count)]`. Former will just make it not run unless we run it manually, latter will cause it to retry up to a few times in case it fails. Can also mark it `[Ignore("Reason goes here.")]`

PR #735, opened 2025-12-29
Add a new aura for oracles.

moonheart08:
## About the PR
This allows Empaths to identify each-other and other oracle-like roles, giving them a stronger base to build trust on.

## Why / Balance
Information economy in ES is currently very gimped, making social deduction harder than it should be. This gives oracles some extra info (who else is an oracle, without room for jesters or future not-seen-as-murderous traitors to interfere) and should help them band together.

## Media
image

## Remarks
Tested this on a few masks, both oracles (empath) and non-oracles, all seems to be working.


K-Dynamic:
Aura farming

PR #736, opened 2025-12-29
New tips dataset

mirrorcult:
## About the PR

just pruned version of the old one (removed obviously irrelevant stuff and modified text on some)

commits are mostly structured sanely, see 79b581c7a9869c137a46eabe9c7ba3696baafe47 for an actual diff between upstream tips and the new one (except a couple ones i removed later)


mirrorcult:
needs a lot of new tips and theres a lot to write for those but leaving it for later

PR #739, opened 2025-12-30
Scale viewcone occludable alpha correctly

mirrorcult:
## About the PR

i intuited this bugs existence from something that wasnt actually the bug and idk what this would actually affect thats ingame right now at all. but its fixed


Princess-Cheeseballs:
Does this fix the viewcone messing with the damage overlay or is that a different bug?

mirrorcult:
different bug

PR #743, opened 2025-12-31
Subverter chip improvements

### LINKED ISSUE #686: subverter chip has questionable ux
it really should display how to use it (click on crit crew member) in the examine text. it has popups but it could do more.

### LINKED ISSUE #741: subverter chips shouldnt fail on non-syndie rounds
we have the technology. It Just Works

EmoGarbage404:
## About the PR

Closes #686 
Closes #741 

lightly modifies the chip description to explain how to use it. Oops.

Changes the conversion logic so that it starts the rule if one doesn't exist. This means that if you use a subverter chip in a non-traitor round, it'll just start up the traitor rule and let them play as if it was one. Extremely funny.

Additionally fixes an extremely minor bug: in the event that the mask is added without a proper troupe entity, the troupe would roll objectives as if there were no members. This means that you could get a kill objective on yourself, which would be pretty damn stupid. Fixed it by just delaying when we "start" the rule until after we've added the player onto the troupe.

image

### DIFF
--- a/Content.Server/_ES/Masks/ESMaskSystem.cs
+++ b/Content.Server/_ES/Masks/ESMaskSystem.cs
@@ -1,6 +1,4 @@
-using System.Diagnostics.CodeAnalysis;
 using System.Linq;
-using Content.Server._ES.Auditions;
 using Content.Server.Chat.Managers;
 using Content.Server.GameTicking;
 using Content.Server.Roles.Jobs;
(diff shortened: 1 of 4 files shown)


Content.Server/_ES/Masks/ESMaskSystem.cs
```diff
@@ -174,13 +169,13 @@ public override void ApplyMask(Entity<MindComponent> mind, ProtoId<ESMaskPrototy
     {
         var mask = PrototypeManager.Index(maskId);
 
-        if (troupe is null)
+        // If we are spawning a new rule, we should initialize the objectives *after*
+        // the first player is added to ensure targeting shenanigans don't happen.
```
moonheart08, line 173: If this is the case, masquerades currently instantly start the rule during mask assignment. This likely needs fixed, i'll try to remember to get to it.

Issue #802, opened 2026-01-04
Traitors, RandomTraitors, and RedCarpet need a minor difficulty curve tune


moonheart08:
They need a third traitor sooner than 18 players. Perhaps at 12?

EmoGarbage404:
iirc in the original setup the third traitor would show up at 13, the next at 19, and the fifth at 25

EmoGarbage404:
i think its been done reasonably well enough

PR #814, opened 2026-01-04
Masquerade balance tweaks

moonheart08:
## About the PR
Adjusts tot balance a bit for #802. Experimentally ensures some of the "power tots".

## Media

N/A


EmoGarbage404:
I'm kind of suspect of the second set of power traitors that a lot of the masquerades pick up around ~25 pop. Having 2 subverters is an extra 4 recruits and part of the design of demolitionist specifically mentions limiting it to 1 per round for balancing reasons. I think it'd probably be best to delay the confirmed power traitor until closer to 20 pop and then keep the other traitors as confirmed non-power traitors (unless the last one is like AllTraitors with a low chance of rolling subverter or demo)

moonheart08:
Changes made.

PR #817, opened 2026-01-05
Sleepy Pen Changes

### LINKED ISSUE #809: The Assassin's sleepy pens should be turned in to medipens
More intuitive, more plausibly deniable, and would make the item feel way better to use in general.

Princess-Cheeseballs:
## About the PR

Sleepy pen now looks and behaves like a medipen with the caveat that interacting in hand doesn't have you inject yourself. 

## Media

N/A


EmoGarbage404:
image

PR #828, opened 2026-01-06
Server, wooden door, bookshelf, and crayon box sprites

Flareguy:
## About the PR
Resprites the crew monitoring server, wooden door, crayon box, and bookshelf (partially)

## Media
Content Client_MA0WF1N0O1


EmoGarbage404:
>telecoms
erm its a server ☝️🤓 

also i think that the crew monitoring server should keep some kind of identifiable med marking so we can potentially reuse the base server for other stuff in the future.

K-Dynamic:
I think putting the coloured boxes on the side communicates less info compared to leaving them on the front. The crew monitor in the screenshot is basically indistinguishable from other servers.

mirrorcult:
its not great that the actual sprite info of the server is obscured by the thindow rn

mirrorcult (changes requested):
crewmon server light panel should be on the front face instead of the side for readability, rn its literally just obscured where its mapped on toast

mirrorcult (approved):

PR #830, opened 2026-01-07
Disallow non-tanks in Suit Storage slot

### LINKED ISSUE #805: Non-tanks shouldnt go into suit storage
fuck that noise

thesaladmancer:
## About the PR

Removes `SuitStorage` from allowed inventory slots of most (if not, all) non-tank items, allowing only internals tanks to be equipped on the suit storage slot.

Initially I was going to also remove the suit storage slot from non-suit outer clothing items, but I feel that may be better for a separate PR (if that is even desired after restricting the inventory slot to only tanks.)

Resolves #805.

## Media

N/A


K-Dynamic:
Change the suit storage icon so it looks like a tank

PR #833, opened 2026-01-08
Add statues

Sadie-silly:
im so tired. i want to go to bed. i made this PR first goodnight

image


Sadie-silly:
should have fixed test fail, double check though because i lazily webedited it and dunno if the formatting or anyhting broke

EmoGarbage404 (changes requested):
proto IDs should be prefixed with "ES"

Sadie-silly:
> proto IDs should be prefixed with "ES"

oops, fixed

PR #835, opened 2026-01-08
minor gun balance changes

Flareguy:
## About the PR
tweaks some gun stats to address minor balancing concerns.

- riot shotguns shoot slower and have a much wider spread, making them a lot more specialized. the lower fire rate might feel pretty bad but we are going to learn if that is the case in about 11 hours
- the energy rifle now does a lot more damage and has slightly faster projectile speed. it was not very good before but should be very energy-efficient per mag hopefully
- the caps laser shoots a volley of 3 lasers instead of 2 to make it feel more unique, and can shoot more lasers overall to compensate

## Media
[Content.Client_sH0nzDYfCa.webm](https://github.com/user-attachments/assets/3ff04ee2-59ce-4db8-9954-cd1be86d54da)

### DIFF
--- a/Resources/Prototypes/_ES/Entities/Objects/Weapons/Guns/Projectiles/energy.yml
+++ b/Resources/Prototypes/_ES/Entities/Objects/Weapons/Guns/Projectiles/energy.yml
@@ -44,7 +44,7 @@
       path: /Audio/_ES/Weapons/Guns/Hits/sear.ogg
     damage:
       types:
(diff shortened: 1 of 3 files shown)


EmoGarbage404 (commented):
did you mean to adjust gunspreadmodifier for the double barrel as well?

Flareguy:
nope. pushed a fix. oops

mirrorcult (approved):

PR #849, opened 2026-01-09
chloral/knockout medipen buff

### LINKED ISSUE #848: knockout medipen should keep people asleep for longer
it's like a 15 second drowsy timer to only knock someone out for like 10 seconds

Flareguy:
## About the PR
Buffs chloral hydrate in order to make the knockout medipen better. It now uses ForcedSleep instead of drowsiness, meaning that you will go to sleep after a fixed amount of time & not be able to just wake up randomly.

The fixed sleep seems to last about 28~ seconds with current values at 10u.

## Media
[Content.Client_y0ohQZOWhO.webm](https://github.com/user-attachments/assets/123c8983-63c5-4d5a-80d5-4a1ee1b19a0d)

### DIFF
--- a/Resources/Prototypes/Reagents/toxins.yml
+++ b/Resources/Prototypes/Reagents/toxins.yml
@@ -68,10 +68,12 @@
       - !type:MovementSpeedModifier
         walkSpeedModifier: 0.65
         sprintSpeedModifier: 0.65
+# ES Start
       - !type:ModifyStatusEffect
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Reagents/toxins.yml
```diff
@@ -81,6 +83,7 @@
           types:
             Poison: 1.5
 
+
```
EmoGarbage404: ominous newline of doom

PR #853, opened 2026-01-09
Stagehand chat notification system

mirrorcult:
## About the PR

- generic thing for sending stagehands messages about stuf that happens in round
- handlers for kill reporting + objective completion/failing yaay

sus things
- querying all handlers to get the one that owns an objective. its weird but uhh theres not that many of them and it only happens when an objective for sure completes/fails, not just when progress is updated at all, so its not too bad? we dont store this data anywhere else i think
- removed access from objectivehodler im sorry this shit doesnt do anything and doesnt help
- had to give the gamerules names since they are the objective holders. pretty sure thats fine

## Media

image
image

### DIFF
--- a/Content.Server/_ES/Stagehand/ESStagehandNotificationsSystem.cs
+++ b/Content.Server/_ES/Stagehand/ESStagehandNotificationsSystem.cs
@@ -1,8 +1,8 @@
 using Content.Server.Chat.Managers;
 using Content.Server.KillTracking;
+using Content.Shared._ES.Objectives.Components;
 using Content.Shared._ES.Stagehand.Components;
 using Content.Shared.Chat;
-using Content.Shared.Players;
 using JetBrains.Annotations;
 using Robust.Server.Player;
 using Robust.Shared.Network;
@@ -23,14 +23,16 @@ public sealed class ESStagehandNotificationsSystem : EntitySystem
         base.Initialize();
 
         SubscribeLocalEvent<KillReportedEvent>(OnKillReported);
+        SubscribeLocalEvent<ESObjectiveProgressChangedEvent>(OnObjectiveProgressChanged);
     }
 
-    private void OnKillReported(KillReportedEvent ev)
+    private void OnKillReported(ref KillReportedEvent ev)
     {
         if (!TryComp<ActorComponent>(ev.Entity, out var actor))
             return;
 
         string? msg = null;
+        var severity = ESStagehandNotificationSeverity.Medium;
 
         if (ev.Suicide)
         {
@@ -56,6 +58,7 @@ public sealed class ESStagehandNotificationsSystem : EntitySystem
             if (!_player.TryGetSessionById(player.PlayerId, out var attackerSession) || attackerSession.AttachedEntity is not { } attackerEnt)
                 return;
 
+            severity = ESStagehandNotificationSeverity.High;
             msg = Loc.GetString("es-stagehand-notification-kill-player",
                 ("entity", ev.Entity),
                 ("username", actor.PlayerSession.Name),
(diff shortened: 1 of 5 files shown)


Content.Shared/_ES/Objectives/Components/ESObjectiveHolderComponent.cs
```diff
@@ -6,7 +6,6 @@ namespace Content.Shared._ES.Objectives.Components;
 /// Denotes an entity that can have objectives associated with it.
 /// </summary>
 [RegisterComponent, NetworkedComponent, AutoGenerateComponentState(true)]
-[Access(typeof(ESSharedObjectiveSystem), Other = AccessPermissions.None)]
```
EmoGarbage404, line 9: i dont mind if people read off these since they're cached and always have up-to-date info, but it should at least have write restrictions on it.

Content.Server/_ES/Stagehand/ESStagehandNotificationsSystem.cs
```diff
@@ -0,0 +1,150 @@
...
+            if (!holder.OwnedObjectives.Contains(ev.Objective.Owner))
+                continue;
+
+            // this assumes only one holder can own an objective, which I think is true right now, and the purpose of owned objectives anyway?
+            foundHolder = uid;
+            break;
+        }
+
+        if (foundHolder is null)
+            return;
```
EmoGarbage404, line 106: this should be pulled out into objective system as like TryGetObjectiveOwner

that would also remove the need for the access changes to the component

EmoGarbage404 (changes requested):

Content.Shared/_ES/Objectives/Components/ESObjectiveHolderComponent.cs
mirrorcult: write restrictions don't function like that on collections, to be able to do `.Contains` you need execute perms and having execute perms lets you add and remove stuff from the collection since write perms just handles assignment

mirrorcult: write restrictions don't function like that on collections, to be able to do `.Contains` you need execute perms and having execute perms lets you add and remove stuff from the collection since write perms just handles assignment

mirrorcult:
resolved

Content.Server/_ES/Stagehand/ESStagehandNotificationsSystem.cs
```diff
@@ -0,0 +1,140 @@
...
+        else if (ev.Primary is KillNpcSource npc)
+        {
+            msg = Loc.GetString("es-stagehand-notification-kill-npc",
+                ("entity", ev.Entity),
+                ("username", actor.PlayerSession.Name),
+                ("attacker", npc.NpcEnt));
+        }
+        else if (ev.Primary is KillPlayerSource player)
+        {
+            if (!_player.TryGetSessionById(player.PlayerId, out var attackerSession) || attackerSession.AttachedEntity is not { } attackerEnt)
```
moonheart08: Not reporting the death at all because someone lost connection just after doing it sounds undesirable.

moonheart08 (changes requested):

Content.Server/_ES/Stagehand/ESStagehandNotificationsSystem.cs
mirrorcult: its mostly an issue with kill reporting rn where it doesnt actually store any info about player sources except their netid, so theres not much info we can get out of it

that said i guess might as well just report it as "player was killed by somebody idk lol"

moonheart08 (approved):

PR #856, opened 2026-01-10
Show mask and troupe descriptions in character menu

mirrorcult:
## About the PR

feedback shows ppl dont rly understand what their masks are intended to do always just from context

- character menu is sized bigger by default (more space to show this stuff)
- mask descs rewritten
- troupe descs added

## Media

image


EmoGarbage404:
it's kinda a lot of unformatted text to just throw on the menu and I feel like having the smaller objectives window is really annoying for usability, since people can have like 4 to 6 objectives at a time.

mirrorcult:
i can just make it even taller by default but i feel like its really necessary to have it there, the conveyance for these is bad right now

EmoGarbage404:
> i can just make it even taller by default but i feel like its really necessary to have it there, the conveyance for these is bad right now

i mean i agree i just feel like it's overly prevalent for information that is only useful once. Like, i would prefer if the troupe/mask names had a little `(?)` icon next to them you could hover over to show the description.

mirrorcult:
image

PR #860, opened 2026-01-10
Open OOC on lobby enter & retheme ooc/deadchat

mirrorcult:
## About the PR

selects ooc channel on enter (feedback)

rethemes ooc chat to 'lobby chat' and deadchat to 'stagehand chat'

## Media

image

### DIFF
--- a/Content.Client/Lobby/LobbyState.cs
+++ b/Content.Client/Lobby/LobbyState.cs
@@ -85,7 +85,9 @@ namespace Content.Client.Lobby
             _contentAudioSystem.LobbySoundtrackChanged += UpdateLobbySoundtrackInfo;
 
             chatController.SetMainChat(true);
+            // ES START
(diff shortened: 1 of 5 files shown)


EmoGarbage404:
i dont mind the stagehand stuff but i feel like lobby chat should probably be it's own thing separate from OOC just because OOC on the backend is like the out-of-sim global messaging and we eventually want lobby chat to be different from that.

also the stagehand chat is like the palest lavender i've ever seen in my life that shit is pasty

mirrorcult:
> i dont mind the stagehand stuff but i feel like lobby chat should probably be it's own thing separate from OOC just because OOC on the backend is like the out-of-sim global messaging and we eventually want lobby chat to be different from that.

i mean yeah but seems out of scope here. i just want the text to be different

> 
> also the stagehand chat is like the palest lavender i've ever seen in my life that shit is pasty

i thought it'd be more lavender but after testing it and it was like off white i realized i kind of liked that

PR #863, opened 2026-01-10
Clues

EmoGarbage404:
## About the PR

Adds clues as a concept and makes them generally accessible

omits some clues from the doc since they need better integration to be usable

## Media

image
image


moonheart08 (approved):
rip relationships

EmoGarbage404:
> rip relationships

i casually ripped it out when i was here cuz it was a bunch of unused code that really wasn't set up in a usable way for anything.

PR #889, opened 2026-01-11
Bump lobby timer to 7.5 minutes

### LINKED ISSUE #879: lobby timer should be longer
3 minutes is not long enough. it should be 5 or so

EmoGarbage404:
## About the PR

Closes #879 

when we were doing like 3 rounds a playtest i wanted it short but now that things are rolling smoothly and going longer we need to give people more breathing time for life stuff.

## Media

N/A


mirrorcult:
i think itd be better to be a little higher, like 7.5 minutes or something

PR #892, opened 2026-01-11
Fix some issues with stagehand notifs

### LINKED ISSUE #872: Player kill notification has bad format
Image

damnit

### LINKED ISSUE #875: Stagehadn Mask objective messages say mind (PLAYER NAME) instead of the player's name
only seems to be present for mask objectives.

EmoGarbage404:
## About the PR

Closes #872 
Closes #875 

changes
- fixes the incorrect locale argument passed into the kill notif
- displays the proper name for players when completing mask based objectives
- puts the "!" inside the quotes for objective completion since it looks nicer imo

## Media

N/A

### DIFF
--- a/Content.Server/_ES/Stagehand/ESStagehandNotificationsSystem.cs
+++ b/Content.Server/_ES/Stagehand/ESStagehandNotificationsSystem.cs
@@ -1,5 +1,6 @@
 using Content.Server.Chat.Managers;
 using Content.Server.KillTracking;
+using Content.Shared._ES.Auditions.Components;
 using Content.Shared._ES.Objectives;
 using Content.Shared._ES.Objectives.Components;
 using Content.Shared._ES.Stagehand.Components;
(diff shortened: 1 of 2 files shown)


Content.Server/_ES/Stagehand/ESStagehandNotificationsSystem.cs
```diff
@@ -70,7 +71,7 @@ private void OnKillReported(ref KillReportedEvent ev)
                 msg = Loc.GetString("es-stagehand-notification-kill-player",
                     ("entity", ev.Entity),
                     ("username", actor.PlayerSession.Name),
-                    ("attacker", attackerEnt),
+                    ("playerEnt", attackerEnt),
```
mirrorcult: imo rename the variable in the loc string instead of in c#, that was my intent but i just forgot to edit the loc string

mirrorcult (changes requested):

PR #898, opened 2026-01-12
Fix chameleon prototypes being wack

### LINKED ISSUE #600: Chameleon clothing uses non-ES jobs and clothing options
painful YAML cleanup. Oh boy.

So basically chameleon clothing defaults to using just about everything which is... FUN. This means that a ton of items are in the chameleon clothing pool despite the fact they really shouldn't be there. The controller implant also has job presets for all upstream jobs, most of them with incorrect non-matching outfits.

This needs a massive purge, with all new chameleon clothes prototypes and ideally a custom ES marker component (DONT DO A TAG I WILL KILL YOU) so we can actually curate this a bit

EmoGarbage404:
## About the PR

Closes #600 

Bunch of changes regarding chameleon clothes:
- Removed out chameleon outfits from upstream and replaced them with custom ES ones for jobs
- Removed all non-ES clothing from chameleon clothes. This basically is everything that isn't worn by a job + colored clothing
- Removed the chameleon scarf, mask, and glasses from the bundle. this is because these items are either functionally useless or not worn by meaningful amounts of jobs. (shoutout neck slot, used by only 2 jobs)
- Made the chameleon controlled implant delete on implant
- migrated out a handful of old jumpsuits we weren't using.

## Media

N/A

### DIFF
--- a/Content.Client/Implants/UI/ChameleonControllerMenu.xaml.cs
+++ b/Content.Client/Implants/UI/ChameleonControllerMenu.xaml.cs
@@ -40,7 +40,9 @@ public sealed partial class ChameleonControllerMenu : FancyWindow
         _sprite = _entityManager.System<SpriteSystem>();
         _job = _entityManager.System<JobSystem>();
 
-        _outfits = _prototypeManager.EnumeratePrototypes<ChameleonOutfitPrototype>();
+// ES START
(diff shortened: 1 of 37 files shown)


Sadie-silly:
what two jobs even use the neck slot? I literally can't think of anyone that uses it

EmoGarbage404:
> what two jobs even use the neck slot? I literally can't think of anyone that uses it

sci and det both spawn with ties. they might unironically be the only neck slot items in the game besides maint loot

PR #912, opened 2026-01-13
Nerf autolathe ammo production

### LINKED ISSUE #899: Ammo printing is too strong
honest to god it should be like one bullet at a time and fairly slow and expensive. you can get a ton of ammo so quickly.

having too much ammo degrades resource decay gameplay so we need to be more limiting about it.

thesaladmancer:
## About the PR

Resolves #899.

The autolathe now produces bullets individually, in line with shotgun shells. Magazines and speedloaders produced by the autolathe are also now empty.

Values are as follows:
- Magazines, speedloaders, and all bullets take 5 seconds to produce (standardized with shotgun shells)
- 9mm Magazine (empty) costs 4 steel
- 9mm Round costs 2 steel per bullet
- .357 Speedloader (empty) costs 5 steel
- .357 Round costs 2 steel per bullet

I suspect that the values will be tweaked later anyways, so I didn't put too much deliberation into them for now. However, this is pretty unambiguously intended to make it more difficult for anyone to acquire large amounts of ammo.

## Media

image

### DIFF
--- a/Resources/Prototypes/_ES/Lathe/Recipes/ammunition.yml
+++ b/Resources/Prototypes/_ES/Lathe/Recipes/ammunition.yml
@@ -1,17 +1,37 @@
 # Handgun
 - type: latheRecipe
   parent: BaseAmmoRecipe
-  id: ESMagazine9mm
-  result: ESMagazine9mm
+  id: ESMagazine9mmEmpty
+  result: ESMagazine9mmEmpty
+  icon: # Manually specifies an icon because it needs to show as empty
+    sprite: _ES/Objects/Weapons/Guns/Ammunition/9mm_mag.rsi
+    state: base
   materials:
     Steel: 400
 
 - type: latheRecipe
   parent: BaseAmmoRecipe
-  id: ESSpeedLoader357
-  result: ESSpeedLoader357
+  id: ESSpeedLoader357Empty
+  result: ESSpeedLoader357Empty
+  icon: # Manually specifies an icon because it needs to show as empty
+    sprite: Objects/Weapons/Guns/Ammunition/SpeedLoaders/Magnum/magnum_speed_loader.rsi
+    state: base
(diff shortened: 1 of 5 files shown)


Resources/Prototypes/_ES/Lathe/Recipes/ammunition.yml
```diff
@@ -1,18 +1,38 @@
...
   materials:
     Steel: 400
 
 - type: latheRecipe
   parent: BaseAmmoRecipe
-  id: ESSpeedLoader357
-  result: ESSpeedLoader357
+  id: ESSpeedLoader357Empty
+  result: ESSpeedLoader357Empty
+  icon: # Manually specifies an icon because it needs to show as empty
```
EmoGarbage404, line 18: these empty magazine prototypes should add a `name` field that's like `.357 speedloader (empty)` to clarify that these don't spawn with bullets. I know the icons are empty by default but it should probably be written explicitly so that people don't make them by default.

this also applies to the 9mm mag

Resources/Prototypes/_ES/Lathe/Recipes/ammunition.yml
```diff
@@ -1,18 +1,38 @@
...
+    state: base
   materials:
     Steel: 500
 
+- type: latheRecipe
+  parent: BaseAmmoRecipe
+  id: ESRound9mm
+  result: ESRound9mm
+  materials:
+    Steel: 50
```
EmoGarbage404: I think each individual bullet should still be fairly expensive. I honestly would be okay with keeping them at the same cost as shells and then just experimenting with it later if it ends up too restrictive.

my general philosophy is that i want getting more ammo to be _possible_, just extremely impractical.

EmoGarbage404 (changes requested):
👍

PR #918, opened 2026-01-14
Remove tile pry random offset

### LINKED ISSUE #775: Pried tiles have random rotation when they shouldn't
looks ugly

thesaladmancer:
## About the PR

Resolves #775. We've generally been trying to keep in-world items from having weird rotation, and this is another step in that direction.

## Media

https://github.com/user-attachments/assets/4ff276b5-ee86-49bf-bc69-587ac7b6293e


EmoGarbage404:
i agree with the random rotation being removed but maybe it should keep the random offset on position

thesaladmancer:
After some deliberation with Flare and Emo, I'm gonna close this because it doesn't really look any better than the way it currently looks.

PR #920, opened 2026-01-14
Brush up timers a bit, notably to support not networking them.

moonheart08:
## About the PR
Nothing user facing, makes timers support non-networked usecases and adds tests. Also adds method timers (which take an Action) as a "i really don't care let me do convenience" option.

Also makes an opinionated change: Timer spawn results are now must-use, so you have to consider handling the failure state. This only results in a bunch of discards in this PR, but yknow.

## Media

N/A

### DIFF
--- a/Content.Server/_ES/Mind/ESAutoGhostSystem.cs
+++ b/Content.Server/_ES/Mind/ESAutoGhostSystem.cs
@@ -58,8 +58,11 @@ public sealed class ESAutoGhostSystem : EntitySystem
 
     private void AutoGhost(EntityUid uid)
     {
+        if (!_mind.TryGetMind(uid, out _, out _))
+            return; // Don't ghost the brainless.
+
         // This may fail in some extreme scenarios like a mob state change happening before the entity is even map-init.
         // Just silence it, the failure is fine.
-        _ = _entityTimer.SpawnTimer(uid, AutoGhostDelay, new ESAutoGhostEvent(), logFailure: false);
+        _ = _entityTimer.SpawnTimer(uid, AutoGhostDelay, new ESAutoGhostEvent());
     }
 }
--- a/Content.Server/_ES/Mind/ESAutoGhostSystem.cs
+++ b/Content.Server/_ES/Mind/ESAutoGhostSystem.cs
@@ -58,6 +58,6 @@ public sealed class ESAutoGhostSystem : EntitySystem
 
     private void AutoGhost(EntityUid uid)
     {
-        _entityTimer.SpawnTimer(uid, AutoGhostDelay, new ESAutoGhostEvent());
+        _ = _entityTimer.SpawnTimer(uid, AutoGhostDelay, new ESAutoGhostEvent());
     }
 }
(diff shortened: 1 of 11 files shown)


moonheart08:
this is all amnesiac soap btw

Content.Shared/_ES/Core/Timer/ESEntityTimerSystem.cs
```diff
@@ -12,50 +14,100 @@ namespace Content.Shared._ES.Core.Timer;
...
     [PublicAPI]
-    public Entity<ESEntityTimerComponent>? SpawnTimer(EntityUid target, TimeSpan duration, ESEntityTimerEvent endEvent)
+    [MustUseReturnValue]
+    public Entity<ESEntityTimerComponent>? SpawnTimer(EntityUid target, TimeSpan duration, ESEntityTimerEvent endEvent, bool logFailure = true)
     {
         if (!TimerTargetIsValid(target))
+        {
+            if (logFailure)
+                Log.Error($"Failed to spawn a timer on {target} due to invalidity, event was {endEvent}.");
+
```
mirrorcult, line 67: would prefer if this was more descriptive of the actual reason (terminating/deleted or pre-mapinit)

Content.Shared/_ES/Core/Timer/Components/ESEntityTimerComponent.cs
```diff
@@ -8,14 +8,14 @@ namespace Content.Shared._ES.Core.Timer.Components;
 /// Component that holds data regarding a generic timer.
 /// </summary>
 [RegisterComponent, NetworkedComponent, AutoGenerateComponentState, AutoGenerateComponentPause]
-[Access(typeof(ESEntityTimerSystem), Other = AccessPermissions.None)]
+//[Access(typeof(ESEntityTimerSystem), Other = AccessPermissions.None)] // interferes with integration tests. :(
```
mirrorcult: just remove it instead of commenting it out, i think we probably want to phase out access attrs in EScode more generally for more reasons than just this

Content.Server/_ES/Mind/ESAutoGhostSystem.cs
```diff
@@ -58,6 +58,11 @@ private void OnAutoGhost(Entity<MindContainerComponent> ent, ref ESAutoGhostEven
 
     private void AutoGhost(EntityUid uid)
     {
-        _entityTimer.SpawnTimer(uid, AutoGhostDelay, new ESAutoGhostEvent());
+        if (!_mind.TryGetMind(uid, out _, out _))
+            return; // Don't ghost the brainless.
+
+        // This may fail in some extreme scenarios like a mob state change happening before the entity is even map-init.
+        // Just silence it, the failure is fine.
+        _ = _entityTimer.SpawnTimer(uid, AutoGhostDelay, new ESAutoGhostEvent());
```
mirrorcult: this says silence it but doesnt pass in false for `logFailure`?

mirrorcult (changes requested):
slightly sus on using a null timer event on client as a marker for it being server-only but probably fine?

mirrorcult (approved):
me like

PR #923, opened 2026-01-14
Feat/mask/avenger

EmoGarbage404:
## About the PR

https://ephemeralspace.github.io/docs/design/masks/crew/avenger.html

avengers get a random target they must protect. If their target dies, they must then seek out the person that killed them, using their distance detection ability to close the gap and kill them. Note that they must directly kill them (killtracking) instead of the person just incidentally dying.

for balance i chucked it into softsafes for now but maybe it'd be better served elsewhere.

## Media

image


Content.Server/_ES/Masks/Avenger/ESAvengeOnKillObjectiveSystem.cs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    private void AvengeKill(Entity<ESAvengeOnKillObjectiveComponent> avenge, KillReportedEvent args)
+    {
+        if (!ObjectivesSys.TryFindObjectiveHolder(avenge.Owner, out var holder))
+            return;
+
+        // This isn't really great but eh. it'll do.
+        if (TryComp<MindComponent>(holder, out var mind) &&
+            _player.TryGetSessionById(mind.UserId, out var session))
```
moonheart08, line 46: Death of an SSD player likely won't count?

Resources/Prototypes/_ES/Actions/Masks/avenger.yml
```diff
@@ -0,0 +1,16 @@
+- type: entity
+  parent: BaseMentalAction
+  id: ESActionMaskAvengerSense
+  name: Sense Distance
+  description: Use this on someone to detect their [bold]aura[/bold] and get a sense of their [bold]motives.[/bold]
```
moonheart08: desc is wrong.

Content.Server/_ES/Masks/Avenger/ESAvengeOnKillObjectiveSystem.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            var msg = Loc.GetString(locale, ("name", name), ("location", locationString));
+            var wrappedMsg = Loc.GetString("chat-manager-server-wrap-message", ("message", msg));
+            _chatManager.ChatMessageToOne(ChatChannel.Server, msg, wrappedMsg, default, false, session.Channel, Color.Red);
+        }
+
+        // Check for
+        // - if the victim killed themselves
+        // - if the victim wasn't killed by a player
+        // - if the killer was actually the person who had to protect this person.
```
moonheart08, line 67: wouldn't it be funnier if they had to kill themselves for this

moonheart08 (changes requested):

Content.Server/_ES/Masks/Avenger/ESAvengeOnKillObjectiveSystem.cs
EmoGarbage404: This only controls the message being sent in chat, which we cannot send if the player is disconnected. The logic will still work as intended wrt objectives and whatnot.

Content.Server/_ES/Masks/Avenger/ESAvengeOnKillObjectiveSystem.cs
EmoGarbage404: Maybe? I just think it's more confusing since you might not realize who the killer is immediately.

moonheart08 (approved):

PR #927, opened 2026-01-15
Tests + Testframe improvements

moonheart08:
## About the PR
Improves the testing frame a bit and adds some extra tests.

### DIFF
--- a/Content.IntegrationTests/Tests/_ES/Masks/MaskTests.cs
+++ b/Content.IntegrationTests/Tests/_ES/Masks/MaskTests.cs
@@ -2,6 +2,7 @@ using Content.IntegrationTests.Tests._Citadel;
 using Content.IntegrationTests.Tests._Citadel.Constraints;
 using Content.Server._ES.Masks;
 using Content.Server.Mind;
+using Content.Shared._ES.Masks;
 using Content.Shared._ES.Masks.Components;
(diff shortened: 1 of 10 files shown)


Content.IntegrationTests/Tests/_ES/Masks/MaskTests.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    public override PoolSettings PoolSettings { get; } = new()
+    {
+        Dirty = true,
+        Connected = true, // We need a guy to mask up.
+    };
+
+    // I solemnly swear this is better than iterating the prototypes.
+    public static string[] Masks =
+    {
```
EmoGarbage404: I feel like this should 100% iterate the prototypes. Expecting people to update this for every new mask isn't reasonable.

moonheart08: While I do agree, doing it as it currently does allows you to run mask tests individually and thus see their failures individually, and there's no way to gather this data from the game early enough for the test framework to be happy or I'd be doing so.

tl;dr it's a big qol boost both for running the tests and also reading test fail logs.

It's a bit of a tricky spot, but I can change it.

moonheart08:
should be good to go.

PR #937, opened 2026-01-17
C4 Rebalance

### LINKED ISSUE #936: Differentiate C4 from explosive grenades
C4 is missing most of its unique interaction stuff from 13 which we want to have to help distinguish it from 'nades. basically, the C4 is pretty much just a normal offensive grenade tool with the additional possibility of sticking it, which just makes it very strong. we want the following changes made to it:

- Don't allow priming it by hand. It should need to be stuck to a wall in order to activate.
- The explosion should be pretty small. C4 is a single-wall breaching tool and should not be balanced with that in mind. no big 4x4 explosion or whatever
- C4 should only blow up like one wall and not really much else. if they're normal solid walls it can knock 'em into girders but none of this "goodbye wall" stuff.

Ermucat:
## About the PR

Made C4 no longer activate unless you stick it to something and turned it into a small one tile explosion
## Media

https://github.com/user-attachments/assets/4b52cd48-5b84-4a76-a4e7-2bff96d2cbbe

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Bombs/plastic.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Bombs/plastic.yml
@@ -54,8 +54,10 @@
     quickEquip: false
     equipDelay: 3
     unequipDelay: 6
-  - type: TriggerOnActivate
-  - type: TriggerOnSignal
+  # ES CHange - only activated by stuck
+  #- type: TriggerOnActivate
+  #- type: TriggerOnSignal
+  # ES End
   - type: TriggerOnStuck
     keyOut: stuck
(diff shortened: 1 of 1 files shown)


EmoGarbage404:
The damage should definitely still be enough to destroy a single wall or reinforced wall. I think i was unclear in my issue: it's fine if it breaks adjacent solid walls, but it should always consistently clear out a single wall that it's planted on. otherwise, it's too weak to be useful compared to the EMAG.

Ermucat:
Boosted damage a good bit, it now breaks reinforced walls and adjacent walls

EmoGarbage404:
i tested the new values locally and the way the damage hits adjacent tiles is really strange but i dont have a better solution off the cuff so i think this'll do for now

PR #954, opened 2026-01-18
Make ID computer ignore non ES jobs

jessicamaybe:
## About the PR

ID Computer now only shows jobs prefixed with ES

## Media

image


EmoGarbage404:
I kinda hate how this works but I guess it's fine to do it like this until this stuff gets properly pulled out into a component or smth.

PR #959, opened 2026-01-19
Atmos spacing visuals

mirrorcult:
## About the PR

o ya ya o ya

used old watervapor sprite for this cuz it looks pretty much like what i want anyway

points in the direction of pressure movement, so vaguely points towards the source of the spacing, which is cool

## Media

https://github.com/user-attachments/assets/e18d69e9-8aa5-4549-a673-af6513687030

### DIFF
--- a/Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
+++ b/Content.Server/Atmos/EntitySystems/AtmosphereSystem.HighPressureDelta.cs
@@ -17,7 +17,10 @@ namespace Content.Server.Atmos.EntitySystems
     {
         private static readonly ProtoId<SoundCollectionPrototype> DefaultSpaceWindSounds = "SpaceWind";
 
+        // ES START
+        // 75 -> 60
         private const int SpaceWindSoundCooldownCycles = 60;
+        // ES END
 
         // ES START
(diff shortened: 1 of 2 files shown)


moonheart08 (approved):

EmoGarbage404:
I almost want there to be more of the particle spawns. the video has a sizeable gap and the actual distribution of the effect is pretty sparse to a point where I think you'd probably just see the actual burst in the wall before actually seeing the effects.

ArtisticRoomba:
might want to move this into the `if (tile.PressureDifference > 100)` branch as otherwise you're waiting on the sound cooldown

mirrorcult:
i prefer waiting on the same cooldown as sounds because i like that the sound and visuals are linked in some way

mirrorcult:
> I almost want there to be more of the particle spawns. the video has a sizeable gap and the actual distribution of the effect is pretty sparse to a point where I think you'd probably just see the actual burst in the wall before actually seeing the effects.

i had it artificially lowered randomly but i can just remove that i suppose. and maybe reduce the cooldown a bit

mirrorcult:
removed the chance to not spawn a vfx + reduced the cooldown in general

EmoGarbage404:
still was kinda mild in testing but it's a good improvement

PR #969, opened 2026-01-20
misc. job yaml changes

Flareguy:
## About the PR
Gives engineers and doctors full access to their department because there's really no excuse for it to not be like that.
Removes combat gloves from the HOS.
Removes janitor rubbergloves because they are ugly and replaces them with cool orange sanitary gloves.
Removes the industrial backpack from atmos techs because honestly the grey backpack looks way better on them.

## Media
image

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Hands/colored.yml
+++ b/Resources/Prototypes/Entities/Clothing/Hands/colored.yml
@@ -71,7 +71,7 @@
   parent: ClothingHandsGlovesSyntheticBase
   id: ClothingHandsGlovesColorBlue
 # ES Start
-# Turned in to nitrile glovesrange
+# Turned in to nitrile gloves
   name: nitrile gloves
   description: High-quality nitrile gloves. Expensive medical PPE.
   components:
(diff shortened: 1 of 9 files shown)


Resources/Prototypes/Entities/Clothing/Hands/colored.yml
```diff
@@ -71,7 +71,7 @@
   parent: ClothingHandsGlovesSyntheticBase
   id: ClothingHandsGlovesColorBlue
 # ES Start
-# Turned in to nitrile gloves
+# Turned in to nitrile glovesrange
```
mirrorcult: huh

Resources/Prototypes/Entities/Clothing/Hands/colored.yml
```diff
@@ -262,8 +262,10 @@
 - type: entity
   parent: ClothingHandsGlovesSyntheticBase
   id: ClothingHandsGlovesColorOrange
-  name: orange gloves
-  description: Regular orange gloves that do not keep you from frying.
+# ES Start
+  name: sanitation gloves
+  description: Biohazard-resistant orange gloves. Useful for preventing injuries from chemical spills.
```
mirrorcult, line 267: feel like these should be their own proto instead of an edit

mirrorcult (changes requested):

PR #1028, opened 2026-01-25
Widely push empath out of masquerades pending rework.

### LINKED ISSUE #1023: remove empath
IMO oracles that give you very direct knowledge are kind of fundamentally maligned with how deduction works in this game. they have no good counterplay other then players going guns blazing pre-emptively and feel really bad to be on the recieving end of.

empath in particular bypasses the sandbox completely, is on a super easy-to-use ability, and gives you a nigh-bulletproof case of who the bad guy is if you know what to look for. and in a non-fixable way since 'bad guy + traitor gear' is always going to be a smoking gun unless you make traitor caches so common as to make actually confirming someone as anything really hard.

moonheart08:
## About the PR
I just like insider more.

When reintroducing empath they should likely be a higher pop role around the 24 player mark, replacing the insider there.

Effectively resolves #1023

## Media

N/A

### DIFF
--- a/Resources/Prototypes/_ES/Masquerades/redcarpet.yml
+++ b/Resources/Prototypes/_ES/Masquerades/redcarpet.yml
@@ -15,7 +15,7 @@
       10: ["ESPhantom"]
       11: ["ESVIP"]
       13: ["ESAssassin"]  # 0 generic crew, 3 traitors.
-      14: ["ESEmpath"]
+      14: ["ESInsider"]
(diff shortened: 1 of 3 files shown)


EmoGarbage404 (changes requested):
I'd prefer if they were just entirely removed since I think they're gonna need a fairly substantive rework + ideological reframing of what we want oracle gameplay to look like as a whole.

moonheart08:
Closing in favor of #1033

PR #1029, opened 2026-01-25
Armor DT + minor doublebarrel nerf

Flareguy:
## About the PR
Replaces all existing armor with DT variants. these values have not been thoroughly messed with but are probably more then fine enough.
Also removes riot armor from the armory & lowers the amount of shotguns available.

Also makes the DB not have insane firerate because being able to instakill people is probably not good.

## Media
video too big but it is tested. trust

### DIFF
--- a/Resources/Prototypes/_ES/Catalog/Tables/job_lockers.yml
+++ b/Resources/Prototypes/_ES/Catalog/Tables/job_lockers.yml
@@ -7,7 +7,7 @@
     - id: CigarGoldCase
       prob: 0.25
     - id: ClothingHeadsetAltCommand
-    - id: ClothingOuterArmorCaptainCarapace
+    - id: ESClothingOuterArmorCaptainCarapace
     - id: MedalCase
     - id: NukeDisk
     - id: PinpointerNuclear
@@ -47,7 +47,7 @@
     - id: ClothingBeltSecurityFilled
     - id: ClothingEyesGlassesSecurity
     - id: ClothingHeadsetAltSecurity
-    - id: ClothingOuterCoatHoSTrench
+    - id: ESClothingOuterCoatHoSTrench
     - id: ClothingShoesBootsJack
     - id: FlashlightSeclite
     - id: RubberStampHos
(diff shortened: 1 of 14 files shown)


Resources/Prototypes/_ES/Catalog/Tables/job_lockers.yml
```diff
@@ -95,7 +95,7 @@
     - id: ClothingHeadHatDetGadget
       prob: 0.15
     - id: ClothingNeckTieDet
-    - id: ClothingOuterVestDetective
+    - id: ESWeaponHybridTaser
```
EmoGarbage404, line 98: i feel like not giving the det any kind of armor is a bit crazy. the taser addition is whatever but they should at least have some kind of armor even if it's just spriteswapped sec armor.

Resources/Prototypes/_ES/Entities/Clothing/OuterClothing/armor.yml
```diff
@@ -0,0 +1,86 @@
...
+  - type: Clothing
+    sprite: Clothing/OuterClothing/Armor/security.rsi
+  - type: Armor
+    modifiers:
+      flatReductions:
+        Blunt: 5
+        Slash: 5
+        Piercing: 5
+        Heat: 5
+        Caustic: 4
```
EmoGarbage404: i know caustic armor and damage is largely an inherited thing but i dont get why we want to preserve it if it's almost entirely useless.

Resources/Prototypes/_ES/Entities/Clothing/OuterClothing/armor.yml
```diff
@@ -0,0 +1,86 @@
+# Base Prototypes
+- type: entity
+  abstract: true
+  parent: [ClothingOuterBaseMedium, AllowSuitStorageClothing]
+  id: ESClothingOuterArmorBase
```
EmoGarbage404, line 5: You missed a bunch of instances of sec armor
- security wardrobes
- marauder caches
- cargo armor crates

EmoGarbage404 (changes requested):
if you want to remove riot armor just remove the entry from the armory spawner. it doesnt need to be manually remapped.

PR #1030, opened 2026-01-25
Add Turncoat

moonheart08:
## About the PR
Turncoat is a mask that can switch sides at will, with the goal to do so strategically and win with either side.

Turncoat starts with the crew in all current masquerades, forcing them to wait til everyone has long left the shuttle to start collecting information on their syndicate counterparts.

Turncoat has not been given a weapon of any kind, may be worth changing this down the line?

Balances them into Traitors and Red Carpet, but not showdown which has less breathing room for the mask and is generally more conflict heavy.

## Media
image
Screenshot_20260125_031849

### DIFF
--- a/Content.Shared/_ES/Masks/ESTargetMaskSystem.cs
+++ b/Content.Shared/_ES/Masks/ESTargetMaskSystem.cs
@@ -1,0 +1,21 @@
+using Content.Shared._ES.Masks.Components;
+using Content.Shared._ES.Objectives.Target.Components;
+
+namespace Content.Shared._ES.Masks;
+
+public sealed class ESTargetMaskSystem : EntitySystem
+{
+    [Dependency] private readonly ESSharedMaskSystem _mask = default!;
+
+    /// <inheritdoc/>
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<ESTargetMaskBlacklistComponent, ESValidateObjectiveTargetCandidates>(Handler);
+    }
+
+    private void Handler(Entity<ESTargetMaskBlacklistComponent> ent, ref ESValidateObjectiveTargetCandidates args)
+    {
+        if (_mask.GetMaskOrNull(args.Candidate) is {} mask && ent.Comp.MaskBlacklist.Contains(mask))
+            args.Invalidate();
+    }
+}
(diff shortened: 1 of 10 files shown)


SuperGDPWYL:
I'm gonna say I personally don't like the Turncoat knowing who the traitors are. ~~I actually really like how they don't know who the kill target is though.~~ nevermind that's just when there's no valid target

I imagine the cases where the Turncoat sides crew are going to feel even worse as traitor than dealing with an Empath. They could be lying but it puts heavy scrutiny on them due to no fault of their own.

Content.Shared/_ES/Masks/MaskCycle/ESActionChangeMaskSystem.cs
```diff
@@ -0,0 +1,32 @@
...
+    [Dependency] private readonly ESSharedMaskSystem _mask = default!;
+    [Dependency] private readonly SharedMindSystem _mind = default!;
+
+    /// <inheritdoc/>
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<ESActionChangeMaskEvent>(Handler);
+    }
+
+    private void Handler(ESActionChangeMaskEvent args)
```
EmoGarbage404, line 19: calling this "handler" instead of OnX is a little bit crazy

Content.Shared/_ES/Masks/ESTargetMaskSystem.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    [Dependency] private readonly ESSharedMaskSystem _mask = default!;
+
+    /// <inheritdoc/>
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<ESTargetMaskBlacklistComponent, ESValidateObjectiveTargetCandidates>(Handler);
+    }
+
+    private void Handler(Entity<ESTargetMaskBlacklistComponent> ent, ref ESValidateObjectiveTargetCandidates args)
```
EmoGarbage404, line 16: ditto

EmoGarbage404 (changes requested):
I think the turncoats being able to see all traitors makes them extremely crew sided in a hard-to-counter way. Becoming a traitor, noting who all other traitors are, then reporting them to sec is pretty easy and near-impossible to counter. I'd rather they just have the objectives of traitors and be able to contribute to them but are otherwise not really able to collaborate with syndies.

moonheart08:
if we want this guy i'll probably just remake 'em.

PR #1031, opened 2026-01-25
Add Syndie Superfan

moonheart08:
## About the PR
The superfan is a mask that converts if the traitor troupe would die out, becoming a new traitor specified by the masquerade.

Depends on #1030, probably conflicts with #1027 and #1028.

## Unanswered questions
- [ ] The superfan needs to be visible to the tots, how would I make the syndie icon exist on them for the traitors, but also not reveal the traitors to them?

## Media
Screenshot_20260125_044657

Screenshot_20260125_044726


EmoGarbage404 (commented):
I think the superfan config should definitely just be on the mask targeting a troupe rather than on a subset of the troupe. Having a subtle variation like that is probably gonna make the mask really strange in subtle ways between different masquerades. I'd much rather they consistently convert on all traitors dead and perhaps we could add another mask that comes in if some guy dies. (generic mask that's like 'backup?' if someone dies you inherit their objectives?)

> The superfan needs to be visible to the tots, how would I make the syndie icon exist on them for the traitors, but also not reveal the traitors to them?

behavior already exists. give them the troupe faction icon component on their body and they'll appear as a traitor to other traitors but won't be able to identify others. Troupe icons are asymmetric so only a troupe member can see their troupe icons but having troupe icons doesn't mean you can see them.

Content.Server/_ES/Masks/Superfan/ESSuperfanSystem.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        if (!_masquerade.TryGetMasqueradeData(out var set))
+            return; // Well, no masquerade means no conversion target.
+
+        if (set.SuperfanTarget is not { } entry)
+        {
+            Log.Error("Superfan has no target despite being in a masquerade!");
+            return;
+        }
```
mirrorcult, line 47: seems like this can be reached even if theres no superfan in the masquerade?

moonheart08: It can, do you mind if I just make an issue to write a test for this?
Specifically the test would enumerate the possible set of masks for the masquerade and, if it determines the superfan is reachable, mandate a target.

It should not be reachable in any other way besides forcing it in a masquerade that doesn't have them configured, which I think is valid to consider an error condition but will probably make *other* tests I want to write (like one for an assailant killing every mask and an equivalent for being a bystander to a killing, and asserting no errors occur) fail when they shouldn't.

tl;dr superfan should not be forced in a masquerade that can't handle it

mirrorcult: yeah that would be fine imo

Content.Server/_ES/Masks/Superfan/ESSuperfanSystem.cs
```diff
@@ -0,0 +1,59 @@
...
+            if (!_mind.IsCharacterDeadIc(Comp<MindComponent>(member)))
+                return; // Well the troupe ain't dead.
+        }
+
+        if (!_masquerade.TryGetMasqueradeData(out var set))
+            return; // Well, no masquerade means no conversion target.
+
+        if (set.SuperfanTarget is not { } entry)
+        {
+            Log.Error("Superfan has no target despite being in a masquerade!");
```
moonheart08:
```suggestion
            // Fail silently, we were never configured to begin with. See #1079
```

PR #1033, opened 2026-01-25
Remove empath

### LINKED ISSUE #1023: remove empath

SuperGDPWYL:
## About the PR
The Empath mask has been removed. All references to it in masquerades have been replaced with an ~~generic Oracle~~ Insider slot. The aura system has been kept for now but is now unused.

## Media
image

### DIFF
--- a/Resources/Prototypes/_ES/Actions/Masks/empath.yml
+++ b/Resources/Prototypes/_ES/Actions/Masks/empath.yml
@@ -46,5 +46,4 @@
   objectiveWhitelist: {}
   priority: 10 # A murderous oracle is still an oracle.
   maskOverrides:
-  - ESEmpath
   - ESInsider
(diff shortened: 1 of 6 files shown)


moonheart08 (changes requested):
All the `#Oracles` set usages should be replaced with explicit insiders.
I'm probably going to remove the oracles set entirely as they're such an important balancing piece i'm not convinced of randomizing them.

EmoGarbage404 (changes requested):
^

SuperGDPWYL:
done

PR #1039, opened 2026-01-26
Antimatter Capsule blows up by C4 now

### LINKED ISSUE #944: C4 isn't strong enough to blow up the antimatter canister
nerf stick swung too hard

YoungThugSS14:
## About the PR

C4 does 60 damage, antimatter capsule now takes 60 damage before blowing up solves https://github.com/EphemeralSpace/ephemeral-space/issues/944
## Media

https://github.com/user-attachments/assets/9f194cae-d137-4148-b4c1-50f182d819eb


EmoGarbage404 (changes requested):
turns out there's a more implicit issue with the prototype. can you change the damage container to StructuralInorganic so that the capsule takes structural damage from the c4? that should make it blow up way easier.

PR #1042, opened 2026-01-26
Make survival boxes smaller

YoungThugSS14:
## About the PR

Survival boxes are 2x2, because right now its 3x3 for no reason
## Media

N/A


EmoGarbage404:
i don't want to make the boxes smaller because eventually i want to make breath masks 1x2 again and i generally feel like the small boxes are optimization slop

YoungThugSS14:
> i don't want to make the boxes smaller because eventually i want to make breath masks 1x2 again and i generally feel like the small boxes are optimization slop

Understand the first part but what does optimization slop mean?

PR #1050, opened 2026-01-29
Knife

### LINKED ISSUE #1046: Kitchen Knife should be replaced with an all purpose knife
### What version did the issue occur in?

_No response_

### Description

The kitchen knife is used in traitor inventories, vendors, etc and should be replaced with a better sprite and to be an all purpose knife.

### Reproduction

N/A

### Additional Context

N/A

EmoGarbage404:
## About the PR

Closes #1046 

Adds a knife. Just a normal knife.

I wanted to have a standard-issue knife to litter all over the game instead of a bunch of domain-specific knives. In general, I want things to be a bit ephemeral and sourceless, in the sense that you can't necessarily pinpoint where a given item came from based off of what it is. In that sense, the knife is more useful cuz there's no longer a distinction of "this is the lathe knife / sec knife / traitor knife"

## Media

image


Sadie-silly:
Knife  e

EmoGarbage404 (commented):
note for myself: buff attackrate to 1.5

PR #1055, opened 2026-02-02
Remove engineering helmet

EmoGarbage404:
## About the PR

With the other suit changes it was annoying me that the engi suit was the only "better" suit in the game. Then i thought about it some more and i realized that i really didn't care about having the embedded flashlight. Or rather, I don't think that the flashlight being in the helmet really contributed anything to the game that wouldn't otherwise be accomplished by having engineers pick up an actual flashlight in-hand. At least that is something that is like a Real Item in the world.

## Media

N/A


Princess-Cheeseballs:
OSHA in shambles

PR #1069, opened 2026-02-05
Removes duplicate chameleon sani gloves

### LINKED ISSUE #987: Two sanitization gloves in chameleon menu
one is the actual sanitization gloves from jani and the other is the identically named orange gloves that have a comp that forces them to appear

DevilishMilk:
## About the PR

Removes the duplicate sanitation gloves from the chameleon menu
Adds the ES suffix to the ES prototype
Fixes https://github.com/EphemeralSpace/ephemeral-space/issues/987

## Media

N/A

### DIFF
--- a/Resources/Prototypes/Entities/Clothing/Hands/colored.yml
+++ b/Resources/Prototypes/Entities/Clothing/Hands/colored.yml
@@ -288,9 +288,6 @@
         color: "#EF8100"
   - type: Fiber
     fiberColor: fibers-orange
-# ES START
-  - type: ESChameleonClothingWhitelist
-# ES END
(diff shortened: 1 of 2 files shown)


Flareguy:
The PR I made like an hour ago already fixed this, along with some other fixes this doesn't do relating to it, like hiding the prototype for the original. Admittedly, I forgot to mark the PR as fixing it, so that's on me.

So, while the gesture is greatly appreciated, I'll have to close this, since the work is already done in a way that requires less fixes then this would require. Outside of it existing & fixing it more thoroughly then this does, this PR would have been fine if not for that.

PR #1074, opened 2026-02-06
Eva access for everyone

PotatOverlord:
## About the PR

Just slapped the access on the ES job proto's
I was gonna make the externals airlock have no access but decided to only touch ES files
Also made the external airlock not RCD resistant because it was for some reason

## Media

N/A

### DIFF
--- a/Resources/Maps/_ES/toast.yml
+++ b/Resources/Maps/_ES/toast.yml
@@ -4035,7 +4035,7 @@ entities:
     - type: Transform
       pos: -1.5,-59.5
       parent: 2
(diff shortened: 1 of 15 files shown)


mirrorcult:
i would really rather the airlocks just be remapped to be unlocked ones. theres an unlocked variant (the base one) and the locked one

PotatOverlord:
Uh idk if this is like good practice for mapping I just changed the proto in the map file and tested it and it worked

mirrorcult:
thats the best way to do it imo yeah

mirrorcult (approved):

PR #1082, opened 2026-02-07
Add hater/secretary

EmoGarbage404:
## About the PR

Adds the hater and secretary masks.

One masks tries to make their target fail their objectives, the other tries to make them succeed their objectives. Both a bit of social-objective fun and i want to see how they roll.

## Media

N/A

### DIFF
--- a/Content.Server/_ES/Masks/Masquerades/ESMasqueradeSystem.cs
+++ b/Content.Server/_ES/Masks/Masquerades/ESMasqueradeSystem.cs
@@ -99,7 +99,8 @@ public sealed partial class ESMasqueradeSystem : GameRuleSystem<ESMasqueradeRule
         var players = ev.Players;
 
         var masksEnum = masks
-            .OrderByDescending(MaskOrder);
+            .OrderBy(m => _proto.Index(m).AssignmentOrder)
+            .ThenByDescending(MaskOrder);
 
         foreach (var mask in masksEnum)
         {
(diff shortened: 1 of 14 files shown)


Content.Server/_ES/Masks/Masquerades/ESMasqueradeSystem.cs
```diff
@@ -98,7 +98,9 @@ private void OnAssignPlayers(ref AssignPlayersToTroupeEvent ev)
 
         var players = ev.Players;
 
-        var masksEnum = masks.OrderByDescending(MaskOrder);
+        var masksEnum = masks
```
moonheart08, line 101: TODO for later: I don't really like masks being assignment order dependent, perhaps a mapinit equivalent for masks down the line so that target assignment can be deferred to when is appropriate.

Content.Shared/_ES/Masks/ESMaskPrototype.cs
```diff
@@ -22,6 +22,12 @@ public sealed partial class ESMaskPrototype : IPrototype, IInheritingPrototype
     [AbstractDataField]
     public bool Abstract { get; private set; }
 
+    /// <summary>
+    /// Arbitray number used to order which masks are assigned before other ones
+    /// </summary>
+    [DataField]
+    public int AssignmentOrder = 10;
```
moonheart08:
```suggestion
    public int AssignmentOrder = 1;
```

moonheart08 (approved):

PR #1117, opened 2026-02-09
Simplify whitelist

### LINKED ISSUE #1108: Simplify whitelist
just always have it on instead of having it disable at highpop its conufinsg

mirrorcult:
## About the PR

closes #1108 

this just makes it basicwhitelist which is just. allow whitelisted players deny nonwhitelisted players

## Media

N/A


mirrorcult:
whoops #1111

PR #1118, opened 2026-02-10
More Aggressive Kill Tracking

EmoGarbage404:
## About the PR

redoes kill tracking to both improve the API and also to make it more aggressive. For gameplay reasons, i basically never want a false negative so the new system treats basically any amount of damage as enough to "kill." Obviously if you just get punched and suicide it won't count but this weights things more towards jesters, which i think is good for gameplay.

## Media

N/A


Content.Shared/_ES/KillTracking/ESKillTrackingSystem.cs
```diff
@@ -0,0 +1,140 @@
...
+        var ev = new ESPlayerKilledEvent(ent, killer);
+        RaiseLocalEvent(ent, ref ev, true);
+    }
+
+    private void AddDamage(Entity<ESKillTrackerComponent> ent, EntityUid? source, FixedPoint2 damage)
+    {
+        if (source.HasValue && !HasComp<MobStateComponent>(source))
+        {
+            // Edge case: sometimes people are gonna pass stupid shit in for the origin
+            // and we don't want inanimate objects counting as kills.
```
moonheart08, line 105: I feel this would be (fish) funny to track at some later (fish) date (fish kill)

moonheart08 (approved):

PR #1125, opened 2026-02-10
Make jesters their own troupe

EmoGarbage404:
## About the PR

Makes all jesters (phantom, martyr, parasite) into their own "jester" troupe separate from crew. 

I wanted to do this for a few reasons:
- Insiders and mechanics relating to troupe identity don't work for jesters, since they count as crew. this makes them very annoying as an element since it reduces their deniability.
- jesters are not really aligned with the crew objective of telescience. At best, they are just not able to help out due to being dead, but at worst, they're actively antagonizing the crew.
- In the future, I'd like a more concrete way to interact with jesters and having their own troupe offers a good way of doing that.

## Media

image

image


Resources/Prototypes/_ES/Masks/masks.yml
```diff
@@ -251,6 +212,46 @@
     all:
     - id: ESObjectiveHaterHate
 
+# Jesters
```
moonheart08: Move them to their own file so we can gradually refactor away from having one big masks file, imo.

moonheart08 (approved):

mirrorcult:
was unsure about doing this initially but yeah im willing to try it out

moonheart08:
test fail is spurious (GHA died)

PR #1141, opened 2026-02-12
Kill shields again 

PotatOverlord:
title


mirrorcult:
id prefer we didnt do this until we have a better way of disabling recipes we dont want (ideally we do something similar to the guidebook where we disable everything and then just dupe the recipes that we want into our own folder)

PR #1179, opened 2026-02-13
Fix stagehand kill announcements

### LINKED ISSUE #1146: kill announcements are mildly borked
I think they're using the wrong entities for some things so everyone is killing themselves ig

EDIT: a suicide didnt announce so i think i messed up the conditions

EmoGarbage404:
## About the PR

Closes #1146 

the variable was wrong but we can also just sub out the logic to make it a bit more sane due to our better API

## Media

image


K-Dynamic:
Imagine a killfeed...

PR #1188, opened 2026-02-14
Add jesters into randomtraitors

### LINKED ISSUE #1153: randomtraitors doesnt have jesters in it
was staring at the config and realized that now that jesters aren't crew, they dont roll in AllCrew so they need to be specified explicitly.

EmoGarbage404:
## About the PR

Closes #1153 

adds jesters back into randomtraitors after they were removed following being split out from crew troupe

## Media

N/A

### DIFF
--- a/Resources/Prototypes/_ES/MaskSets/alljesters.yml
+++ b/Resources/Prototypes/_ES/MaskSets/alljesters.yml
@@ -1,0 +1,4 @@
+- type: esMaskSet
+  id: AllJesters # Every jester, with their mask's weight.
+  maskProvider: !type:ESTroupeMasksProvider
+    troupe: ESJester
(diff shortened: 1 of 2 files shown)


moonheart08:
I feel like this would be better served by making a new mask set that can source random roles from more than one troupe.

EmoGarbage404:
> I feel like this would be better served by making a new mask set that can source random roles from more than one troupe.

My thought was just that random traitors is just defining the "curve" of troupe counts without specifying particular masks, which this is doing well enough. I think trying to do something more random like just picking roles from troupes with weights but just lead to some kinda wacky distributions. And, I think random traitors as a masquerade is balanced well enough so i dont really want to mess with it.

mirrorcult (approved):

PR #1189, opened 2026-02-14
Straight up Lerping it

AftrLite:
And by it, i mean these screenshakes

## About the PR
Adds extra API options to ESScreenshakeSystem for slamming out lerped screenshakes to both PVS and non-PVS filters around designated mapcoordinates yadda yadda yadda

## Media
N/A

### DIFF
--- a/Content.Shared/_ES/Camera/ESScreenshakeSystem.cs
+++ b/Content.Shared/_ES/Camera/ESScreenshakeSystem.cs
@@ -1,6 +1,7 @@
 using System.Linq;
 using System.Numerics;
 using Content.Shared.Camera;
+using Robust.Shared.Map;
 using Robust.Shared.Noise;
 using Robust.Shared.Player;
 using Robust.Shared.Random;
@@ -14,6 +15,7 @@ namespace Content.Shared._ES.Camera;
 /// </summary>
 public sealed class ESScreenshakeSystem : EntitySystem
 {
+    [Dependency] private readonly SharedTransformSystem _transform = default!;
     [Dependency] private readonly IRobustRandom _random = default!;
     [Dependency] private readonly IGameTiming _timing = default!;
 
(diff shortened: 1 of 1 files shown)


moonheart08 (changes requested):
This feels like a kind of odd way to be handling falloff curves, should it not be decaying toward some minimums? Only supporting lerp as your falloff is *fine* for this imo, i looked to see if RT had any generic falloff curve helpers and if it did I'd say use those instead.

The reason I ask what it should be decaying toward is atm this gives them a sharp cutoff between "multiplicatively less effect" and "no effect".

Also not sure lerping *all* the properties actually makes any sense. while the strength of an event falls off with distance, the *duration* (which afaict is controlled by decay) does not.

*Also also* not sure lerp is even the right thing here but it's Probably Fine given the purely visual nature of this and not needing to adhere to the laws of physics.

AftrLite:
Having been extensively tweaking and playing with screenshake values for the better part of a day and a half, i personally found lerping the duration to "feel" better in terms of gamefeel, even if it doesn't neccesarily make sense. 

Regarding sharp cutoffs, Explosions' auditory falloff functions similarly, so i felt it best to keep that "logic". Of course, it wouldn't take me any effort to add even more API options for having a full-to-nothing falloff, rather than a full-to-half cutoff'd falloff, if you feel that'd be appropriate?

AftrLite:
I LOVE OUT-OF-MEMORY BUILDFAILS
THANK YOU GITHUB, VERY COOL
GitHub please give me 1 more crumb of TestRunner RAM memory

mirrorcult (changes requested):
dont like the logic duplication between these 3 functions, shouldnt be that hard to abstract i think?

mirrorcult:
i think ill probably just get this stuff in once i need it for something else later

PR #1193, opened 2026-02-15
Removes shuttle wall rusting

DevilishMilk:
## About the PR

Removes WallReplacementMarker for shuttle walls - just making them not rust at the roundstart and turn into normal walls.
This looked kind of goofy & shuttle walls are probably too Nice to justify making a custom texture for their rust - I can do that if there's interest, however.

## Media

N/A


EmoGarbage404:
I mentioned this in the discord but i kinda dont want to take this approach since this is just a general symptom of the shuttle being a station grid. That's also why things like anomalies spawn on it. It was turned into that so PDA apps worked properly on it but i kinda want to just pursue another fix for that and just make it not a grid, which will prevent all this unwanted variation pass stuff from running on it.

PR #1210, opened 2026-02-16
rename doctor to physician everywhere

### LINKED ISSUE #1121: Physician rename is incomplete
There's still a bunch of things that are named "medical doctor" and refer to the role as such.

PotatOverlord:
## About the PR

title, Idk if changing the names of the image files is a good idea probably not but easy change. I swapped the doctor jumpsuit to an ES prototype. I didn't change any of the tips because we need to completely replace those anyway.
closes #1121 

## Media

N/A


mirrorcult:
the ids and stuff were intentionally not changed, the actual issue is with the upstream entities we use that havent been renamed (i only really renamed ES stuff)

mirrorcult:
like the jumpsuit is still called "medical doctor jumpsuit" because its an upstream thing

sowelipililimute (changes requested):
you can just use .ftl files to provide alternate localized names for these entities
```
ent-ID = physician thingimajig
    .desc = your mum is a physician
```

PotatOverlord:
maybe didn't need to split it into as many ftl files but it might help if we do any similar role changes in the future

mirrorcult (approved):

PR #1216, opened 2026-02-18
Hide, duplicate & prune all construction recipes

mirrorcult:
## About the PR

this is just recipes not graphs or construction comps

- hides all recipes by default
- duplicates all the upstream recipes
- removes all the ones we likely dont want or that i consider bloaty or are not relevant to anything we have at all

## Media

every recipe pictured in grid view. except the fancy black table ignore that ir emoved it after

image
image


EmoGarbage404:
i think i'd want to cut this down a bit more but im happy with this for now.

PR #1218, opened 2026-02-18
Unsmoothed rwalls

mirrorcult:
## About the PR

@PotatOverlord idea

welcome to unsmooth city where the walls are greeb and the lovely pretty

## Media

image


EmoGarbage404:
I'd rather go for some of the midpoint concepts for r-walls we talked about in discord over flatly unsmoothed walls.

PR #1235, opened 2026-02-21
New fonts

mirrorcult:
## About the PR

regular font replaced with tomorrow
fancywindow title font replaced with Geo (will use in more places probably i just like how it looks)

## Media

image


mirrorcult:
hold on no merge yet

PR #1268, opened 2026-02-23
Block jesters from killing people

### LINKED ISSUE #1250: Jesters should be blocked from killing people
use kill tracking and make it so that if a jester kills someone, it duds them and turns them into some mask like "the fool" that otherwise disables their jester abilities and makes them fail their objectives.

EmoGarbage404:
## About the PR

adds some new stuff to jesters. Now, when a jester kills someone, they lose their mask and become The Fool. The fool is just an unaligned jester mask that has an uncompletable objective so they always register as losing.

Closes #1250 

## Media

image
image
image

### DIFF
--- a/Content.Server/_ES/Masks/Jester/Components/ESChangeMaskOnKillObjectiveComponent.cs
+++ b/Content.Server/_ES/Masks/Jester/Components/ESChangeMaskOnKillObjectiveComponent.cs
@@ -10,6 +10,9 @@ namespace Content.Server._ES.Masks.Jester.Components;
 public sealed partial class ESChangeMaskOnKillObjectiveComponent : Component
 {
     [DataField]
+    public LocId Message = "es-fool-conversion-notification";
+
+    [DataField]
     public ProtoId<ESMaskPrototype> Mask;
 
     [DataField]
--- a/Content.Server/_ES/Masks/Jester/Components/ESChangeMaskOnKillObjectiveComponent.cs
+++ b/Content.Server/_ES/Masks/Jester/Components/ESChangeMaskOnKillObjectiveComponent.cs
@@ -12,7 +12,7 @@ public sealed partial class ESChangeMaskOnKillObjectiveComponent : Component
     [DataField]
     public LocId Message = "es-fool-conversion-notification";
 
-    [DataField]
+    [DataField(required: true)]
     public ProtoId<ESMaskPrototype> Mask;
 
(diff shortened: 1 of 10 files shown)


K-Dynamic:
You foool

EmoGarbage404 (commented):
review for myself: add a proper message when you get Dudded

Content.Server/_ES/Masks/Jester/Components/ESChangeMaskOnKillObjectiveComponent.cs
```diff
@@ -0,0 +1,20 @@
...
+/// <summary>
+/// Used for a mask that becomes another mask when they kill someone.
+/// </summary>
+[RegisterComponent]
+public sealed partial class ESChangeMaskOnKillObjectiveComponent : Component
+{
+    [DataField]
+    public LocId Message = "es-fool-conversion-notification";
+
+    [DataField]
```
moonheart08:
```suggestion
    [DataField(required: true)]
```

Content.Shared/_ES/KillTracking/Components/ESKillTrackerComponent.cs
```diff
@@ -54,3 +54,18 @@ public readonly struct ESPlayerKilledEvent(EntityUid killed, EntityUid? killer)
...
+/// Event raised on an entity when they kill and entity with <see cref="ESKillTrackerComponent"/>.
+/// </summary>
+[ByRefEvent]
+public readonly struct ESKilledPlayerEvent(EntityUid killed, EntityUid killer)
+{
+    public readonly EntityUid Killed = killed;
+
+    public readonly EntityUid Killer = killer;
+
+    public bool ValidKill => !Suicide;
```
moonheart08, line 68: Feels subjective enough it shouldn't be here OR should be a method on a system somewhere so it can become more subjective later.

moonheart08 (changes requested):

Content.Shared/_ES/KillTracking/Components/ESKillTrackerComponent.cs
EmoGarbage404: I think that there should probably be a shared interface between the kill report events and then the system can have a method that evaluates them. Rn it's kinda just lazily duped between the two events which is Whatever but obviously not ideal.

PR #1287, opened 2026-02-26
Mask violence test.

moonheart08:
## About the PR
Adds a test to ensure mask-equipped crew members can commit violence w/o crashing.

## Media

N/A


moonheart08:
Stopping here so this doesn't become a thousand+ line PR.

PR #1296, opened 2026-03-03
1984'd Tritium from Plasma

Coolsurf6:
## About the PR
Removed the ability for Plasma to turn into Tritium at high temperatures and supersaturation values. 
Currently the way tritium was being made is now just replaced fully with Carbon Dioxide.

This means that there will no longer be 10000 degree rooms, and gives the crew more of a chance to escape a room that's being plasma flooded. 

Reduced Plasma's heating potential from 160000 to 1000.

Changed the burn rate so plasma burns faster from 9 to a magic number of 0.5 (felt right)

## Media
Screenshot_20260303_201114
Dead (mouse) for scale


moonheart08:
Mark the failing test as `[Explicit]`, it needs rewritten to handle these changes correctly.

PR #1306, opened 2026-03-05
actually fix viewcone draw order

### LINKED ISSUE #393: Viewcones and damage overlay interacts weirdly

mirrorcult:
## About the PR

i might genuinely be stupid

closes #393 

## Media

N/A


Content.Client/_ES/Viewcone/Overlays/ESViewconeConeOverlay.cs
```diff
@@ -35,6 +35,7 @@ public ESViewconeConeOverlay()
     {
         IoCManager.InjectDependencies(this);
         _viewconeShader = _proto.Index(ShaderPrototype).InstanceUnique();
+        ZIndex = -6;
```
moonheart08, line 38: Nit: We should probably just move all these overlays to an *explicit* draw order sometime, defining all the orders in an enum.

Implicit draw order bad.

moonheart08 (approved):

PR #1363, opened 2026-03-08
Offbrand grabbag 2

sowelipililimute:
## About the PR
- co2 retention is gone 🦀 
- brain activity now has a logistic effect on vascular tone rather than a linear one
- brain activity now heals up to a point even without mannitol or inaprovaline
- wounds now passively regenerate at a much higher rate than before if they're below a certain damage threshold, or if they've been tended
- firmcrits no longer make you clumsy
- paincrit now keeps you down
- bic drunkenness is gone


Sadie-silly:
Oh, the bic change <3 i was gojnna make an issue for it and completely forgot but i was drunk almost every round last playtest due to bic

mirrorcult (approved):
neast

PR #1368, opened 2026-03-09
Sprint tweaks

mirrorcult:
## About the PR

- rethemes as adrenaline rush
- some balancing around the numbers, less cooldown, slightly less speed, more duration
- now affects metabolic rate a bit like epi
- now has a comedown which makes you a bit slower with a 5s delay after it ends

## Media

N/A


mirrorcult:
moderately annoying that it didnt show the diffs correctly

Content.Shared/_ES/StatusEffects/Components/ESStatusEffectOnStatusEffectEndedComponent.cs
```diff
@@ -0,0 +1,24 @@
...
+    ///     Status effect to add when ending
+    /// </summary>
+    [DataField(required: true)]
+    public EntProtoId<StatusEffectComponent> StatusEffect = default!;
+
+    [DataField(required: true)]
+    public TimeSpan Duration;
+
+    [DataField]
+    public TimeSpan? Delay;
```
moonheart08, line 30: doc comments, what delay is at a glance isn't intuitive to me and probably won't be to others.

moonheart08 (dismissed):

PR #1411, opened 2026-03-14
Explosion fires

EmoGarbage404:
## About the PR

Explosions now have the ability to start fires.

All default explosions have a 15% chance to start a level-1 fire per each tile. This is pretty ineffective, all things considered. Grenades gets maybe 1-3 fire entities per explosion.

Firebombs and welding fuel tanks have increased flammability properties. They spawn fires 80% of the time and the fires can range from level 1 to 3. For firebombs this gives them an interesting niche as a weak firestarter. Welding fuel tanks have a fire roughly comparable to the electrical fire event which makes them more area denial focused and worse at breaching (how are you going to breach through a fire?)

not really super sanely balanced i just wanted to play around with this as functionality.

## Media

grenade:
image

welding tank:
image

firebomb:
image

### DIFF
--- a/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
+++ b/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
@@ -255,6 +255,17 @@ public sealed partial class ExplosionSystem
                 tileBlocked |= IsBlockingTurf(entity);
             }
         }
+// ES START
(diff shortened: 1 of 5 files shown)


moonheart08 (approved):
I'd like some screenshots mayb. but the code looks good.

EmoGarbage404:
screenshots added

PR #1436, opened 2026-03-15
Overhaul airlock bolt breakage

EmoGarbage404:
## About the PR

Changes the airlock bolt breakage degradation to be user-based rather than airlock-based.

The previous system would just mark the airlocks for degradation, then when they were opened, they'd bolt. This was overwhelming in practice. Primarily, because the airlocks would often be rarely used and thus would never trigger or trigger in obscure low-traffic places.

the new system marks a player, which then causes the next airlock they open (fuzzed a bit) to bolt open. This means that the airlocks affected are more likely to be high traffic zones that have real impact on traversal.

this is also how i originally envisioned the system working anyways. idk why i forgot to implement it like this because the results are like obviously underwhelming.

## Media

N/A

### DIFF
--- a/Content.Shared/_ES/Degradation/ESAirlockFailureSystem.cs
+++ b/Content.Shared/_ES/Degradation/ESAirlockFailureSystem.cs
@@ -1,0 +1,45 @@
+using Content.Shared._Citadel.Utilities;
+using Content.Shared._ES.Degradation.Components;
+using Robust.Shared.Random;
+using Robust.Shared.Timing;
+
+namespace Content.Shared._ES.Degradation;
+
+public sealed class ESAirlockFailureSystem : EntitySystem
+{
+    [Dependency] private readonly IGameTiming _timing = default!;
+
+    /// <summary>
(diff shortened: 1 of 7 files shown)


Content.Shared/_ES/Degradation/ESAirlockFailureSystem.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    /// <summary>
+    /// Consumes an airlock failure charge, decrementing it and cleaning up the component if necessary.
+    /// </summary>
+    public bool TryUseFailureCharge(Entity<ESAirlockFailerComponent?> target)
+    {
+        if (!Resolve(target, ref target.Comp, false))
+            return false;
+
+        var seed = new RngSeed().SeedForStep((int) _timing.CurTick.Value + target.Owner.Id);
```
moonheart08, line 30: So this is how I learn i forgot to make those methods static despite testing them.

im so good at this.

moonheart08 (approved):

PR #1445, opened 2026-03-16
Vigilante Mask

EmoGarbage404:
## About the PR

Implements https://ephemeralspace.github.io/docs/design/masks/crew/vigilante.html

## Media

image


moonheart08 (approved):
lgtm. i do kind of want the kill tracker tally to also count the killed body/mind somehow but that's complicated and out of scope.

PR #1458, opened 2026-03-21
Triple antimatter contact damage

### LINKED ISSUE #1383: Antimatter should kill faster
Due to offmed, it takes around 3 minutes to kill you. This shouldn't be the case.
See #1356 (but probably don't make it do brain damage because that would maybe be a bit weird. idk)

Sadie-silly:
## About the PR
Closes #1383 

Antimatter with 5 dps kills you far too slowly. 15 dps allows you to have a few moments where you accidentally brush against it and live, but still severely punishes you for walking deep inside of it. It still takes a long time to kill you due to offmed, but the injuries you sustain are permanent and will lead to your death eventually.

## Media

https://github.com/user-attachments/assets/748a027c-af59-4039-9902-78fa34a5a971

### DIFF
--- a/Resources/Prototypes/_ES/Entities/Structures/Power/antimatter.yml
+++ b/Resources/Prototypes/_ES/Entities/Structures/Power/antimatter.yml
@@ -37,7 +37,7 @@
   - type: DamageContacts
     damage:
       types:
-        Cellular: 5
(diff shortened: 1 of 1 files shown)


Sadie-silly:
unwanted

PR #1461, opened 2026-03-21
Armory tweaks

### LINKED ISSUE #1423: armory usability issues
- [ ] cooldown is too high. 5 minutes sucks really hard when it fails cuz of a simple mistake
- [ ] the window to press the buttons is too short. It's very hard to tell when a button gets pressed so coordinating it can be a headache. the time pressure being so tight doesn't help.
- [ ] You can't see the cooldown remaining. Should be like a screen or something in the control room.

mirrorcult:
## About the PR

- cooldown timer lowered
- there is now a screen showing the cooldown timer in the room

closes #1423 

## Media

image


mirrorcult:
hold on i didnt actually map it in i just realized

PR #1499, opened 2026-03-24
Adds damage deflection to structures

Flareguy:
## About the PR
adds DD (damage deflection) to basically every structure, which blocks all damage unless it is above the given threshold. a summary of the micro-changes:
- every structure inherited from BaseStructure has 6 DD. you cannot break anything with your fists anymore
- if you do any damage, things will break much faster (at most a minute and a half instead of how it was before where you could do 0.2 damage to a wall for 5 minutes or whatever)
- the days of insane damage resistances for random stuff are gone (no more effectively immortal firelocks and windoors)
- airlocks and everything derivative of them are a lot stronger now and can only really be broken by bullets and the fireaxe
- walls cannot be broken by any weapon in the game. you need to deconstruct them or blow them up
- slash damage is generally now ineffective against most structures (especially windows)
- gunshots deal massively increased damage to windows and computers
- secure containers and blast doors cannot be broken in to easily. even most gunshots will not work, extremely strong weapons are required to break in (like a bunch of .357 rounds or the fireaxe)

this will definitely cause a lot of problems because some stuff randomly doesnt parent basestructure, the relation to damage values and stuff is completely different, etc.
please report any fist-breakable structures you find ingame

## Media
N/A

### DIFF
--- a/Resources/Prototypes/Damage/modifier_sets.yml
+++ b/Resources/Prototypes/Damage/modifier_sets.yml
@@ -1,42 +1,25 @@
 - type: damageModifierSet
   id: Metallic
+# ES Start
+# We need to soft-disable all of the modifier prototypes used by structures, since we have damage deflection for that instead.
   coefficients:
-    Blunt: 0.7
-    Slash: 0.5
-    Piercing: 0.7
-    Shock: 1.2
-    Structural: 0.5
-  flatReductions:
-    Blunt: 5
-    Heat: 5
+    Structural: 1
+# ES End
 
 - type: damageModifierSet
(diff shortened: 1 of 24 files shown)


EmoGarbage404 (changes requested):
the difference between `LockerBase` and `LockerBaseSecure` is that the former is used for non-locking insecure entitystorage like wardrobes. The latter is for things like locked job lockers. I dont want the high resistances applied to both, ideally. It doesnt really matter for normal closets to be weak and i'd rather maintain a distinction between them for balance reasons

PR #1514, opened 2026-03-24
Personal-ID Lockers

### LINKED ISSUE #1474: Personal Lockers
idea starts here: https://discord.com/channels/1381736176351707219/1382262686200631386/1485439059781357568

Basically all the lockers are tied to one person's access/ID whatever which is assigned to the locker when the round starts, and the locker gets an unremovable label on it with their name. Latejoiners just get a locker that isnt assigned yet. (I was inspired a lot by the bay arrival sequence)

This is good because it 1. devalues command's AA superpower 2. population gates a lot of gear that shouldnt be accessible at lower pops and 3. gives everyone a very obvious personal stash that is secure but very suspicous if something is linked to it which is very deduction coin.

Check the discord convo for technical deets

### LINKED ISSUE #1482: AccessListsOriginal gets serialized into maps
Turns out AccessReader component serializes the original access list via lazy loading in a bunch of random places. One of these places is on examine, so anything which gets examined during mapping that has accesses gets this field serialized onto the map.

If you ever change the access of a given entity, this will then disagree with the access on the comp and cause strange lying examine text to show up.

EmoGarbage404:
## About the PR

Closes #1482 
Closes #1474 

Adds lockers which are ID locked. When you spawn you get one assigned to you automatically. All other lockers are just set to "unassigned" by default.

I'd like to eventually pre-assign the lockers to minds before players spawn s.t. every locker has a name on it but uh i cant be fucked to rework all the spawning shit rn.

## Media

image
image


mirrorcult:
itd be nice if it had like a special sprite thing denoting this. like a piece of label paper or something

mirrorcult:
without i think people would just be confused that it doesnt work how they expect

PotatOverlord:
my first thought is making the lock light blue but that seems like colorblindness could mess it up. Bulky card reader that juts out the side maybe?

moonheart08:
Sidenote but I think down the line it'd be Neat if all the possible characters for the round were pre-genned, so even the lockers not yet assigned to a real player still have a name attached and you can wonder "wow who's that guy".

mirrorcult:
iirc that is basically how it works already but its just not really exposed

Content.Server/_ES/Spawning/Components/ESPersonalLockerComponent.cs
```diff
@@ -0,0 +1,18 @@
...
+
+[RegisterComponent]
+[Access(typeof(ESPersonalLockerSystem))]
+public sealed partial class ESPersonalLockerComponent : Component
+{
+    [DataField]
+    public bool Assigned;
+
+    [DataField]
+    public string Name = "unassigned";
```
moonheart08: Make this nullable and look up a loc key when not present, instead.

moonheart08 (changes requested):

mirrorcult (changes requested):
registering my change request that i want some kind of visual difference for personal lockers but otherwise fine

EmoGarbage404:
image

updated with new visuals @mirrorcult

mirrorcult:
theres probably some better way to represent this tbh but looks fine to me

mirrorcult (approved):

PR #1546, opened 2026-03-27
better viewcone grain

mirrorcult:
## About the PR

remembered that the camera static got updated upstream and also remembered that i initially copied this from that shader. so updating it to the new method

## Media

N/A


moonheart08:
just remarkng that I did backport the camera static PR too so we don't have to go fix it.

PR #1552, opened 2026-03-28
viewport hud handling rework + stagehand hud

mirrorcult:
## About the PR

drafting bc i cant figure out why rich text labels are not fucking wrapping

changes how scalingviewport works and fucks around with a bunch of shit until it scales properly how i expect. viewport control now actually takes up the size of the viewport contents

reworks gameplaystate to allow changing screens iwthout needing to change the state and adds a new screen for stagehand hud so it can be diff f rom player hud

turns stagehand vote and observe into controls and properly uicontrollerifies them then embeds them into the stagehand hud

hard to tell how this will affect others with diff setups and resolutions and aspect ratios and stuff idk but in testing it works mostly in the way you would expect

## Media

image

regular play erh ud now
image


mirrorcult:
iseeyoulooking

mirrorcult:
works now with fix from moony inc

PR #1619, opened 2026-03-30
remove botany from toast

Artxmisery:
## About the PR

removed the botany area from toast and divvied its space up between chef and a moved janitor closet
also i ran fixrotations and it hit like things which is why there are 2200 deletions. teehee

## Media
chefbotany


Artxmisery:
ok ignore my fumbling the test fail should go away once #1617 is done

PR #1621, opened 2026-03-31
remove toast botanist job slot

mirrorcult:
## Media

N/A


Artxmisery:
oops thank you for getting that

PR #1627, opened 2026-03-31
viewcone angle modifying

mirrorcult:
## About the PR

i decided raising the event twice (its literally just on client) when the cone renders is better than doing it statefully and having to deal with all that bullshit

no other changes bseides pain status effects + helmet rn we will see how it goes then later

the arrivals station thing is cuz i noticed it was still trying to fucking spawn the arrivals planet map even with the cvar turned off in release. we just need to rip out all the old arrivals code later

## Media

blackout pain (lower pain levels are smaller changes)

image

helmet

image

### DIFF
--- a/Content.Shared/_ES/Viewcone/Components/ESViewconeModifierComponent.cs
+++ b/Content.Shared/_ES/Viewcone/Components/ESViewconeModifierComponent.cs
@@ -1,0 +1,36 @@
+using Content.Shared.Inventory;
+
+namespace Content.Shared._ES.Viewcone.Components;
+
+/// <summary>
+///     Intended to be used on inventory items or status effects (i.e. this is relayed).
+///     Modifies the viewcone angle of the relevant entity additively.
+/// </summary>
+[RegisterComponent, AutoGenerateComponentState]
+public sealed partial class ESViewconeModifierComponent : Component
+{
+    [DataField(required: true), AutoNetworkedField]
+    public float AngleModifier = 0f;
+}
+
+/// <summary>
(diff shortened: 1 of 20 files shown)


Content.Shared/_ES/Viewcone/Components/ESViewconeModifierComponent.cs
```diff
@@ -0,0 +1,36 @@
+using Content.Shared.Inventory;
+
+namespace Content.Shared._ES.Viewcone.Components;
+
+/// <summary>
+///     Intended to be used on inventory items or status effects (i.e. this is relayed).
+///     Modifies the viewcone angle of the relevant entity additively.
+/// </summary>
+[RegisterComponent, AutoGenerateComponentState]
```
EmoGarbage404: probably needs  NetworkedComponent attribute? i lowkey dont know how it works without it.

EmoGarbage404 (dismissed):
would like to see this on the protective suit headpieces but helmet is fine for now

Content.Shared/_ES/Viewcone/Components/ESViewconeModifierComponent.cs
mirrorcult: oh lol whoops

mirrorcult: i feel like that should fail analyzer but

mirrorcult:
image

moonheart08 (approved):
doesn't adjust the welding helm, is that possible?

mirrorcult:
didnt feel like adjusting welding mask here since i think a better solution exists for it (toggling + view distance/semi blindness etc)

PR #1644, opened 2026-04-01
April fools rules tweaks

EmoGarbage404:
## About the PR

Tweak merchant:

- Specify that bans can be mirrored to discord
- Reword rule 1 to remove irrelevant non-rule information
- Change rule 2 to remove "intentionally" specifier
- Reorder terms in rule 3 to be more clear
- Remove mild redundancy in rule 4
- Remove redundant specified in rule 6

## Media

OUT OF DATE:

image

### DIFF
--- a/Resources/ServerInfo/_ES/Guidebook/References/Rules.xml
+++ b/Resources/ServerInfo/_ES/Guidebook/References/Rules.xml
@@ -1,6 +1,6 @@
 <Document>
   The following are the rules for [bold][color=plum]Ephemeral Space[/bold][/color].
-  Failure to follow these rules, knowingly or not, will result in a server ban.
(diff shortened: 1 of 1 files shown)


mirrorcult:
i dont really like the ambiguity of the term 'exploits' in general. it kind of implies that like any unintended glitch use is ban worthy which i dont enjoy

mirrorcult (changes requested):
above

EmoGarbage404:
Reworded rule 4 to "Do not cheat or use modified game clients in any form."

I think this is probably better since it specifies a more broad action (cheating) with a more specific intent (gaining an unfair advantage). I think this would also reasonably cover using exploits to gain an unfair advantage, which is something that i think benefits from being mildly covered in rules.

mirrorcult (approved):

PR #1648, opened 2026-04-02
Delay loading syndiebase map until nuke actually arms

mirrorcult:
## About the PR

adding the troupe anywhere (even in tests) was causing the entire syndiebase map to load even in cases where its obviously never going to be used

this technically lags the server when the nuke arms but in testing it wasnt noticeable and anyway this is just how things should work if the engine was saner and loading maps was actually deferred across ticks and not limited by serializati on code. so i think its the good thing to do

also removes loadmaprule because its only purpose was to preload grids and maps and we really dont want to ever do that

## Media

image


moonheart08 (approved):
this unironically saved 0 seconds of test time somehow but still good change.

PR #1714, opened 2026-04-08
Allow jobs to influence character gen

EmoGarbage404:
## About the PR

changes:
- Pull out all the name generation constants into ESNameConfig. I even added comments!!!
- Shuffle around the player spawning code so that we get the job _before_ the mind
- Generate minds on demand rather than pooling them roundstart
  - Realistically we might want to return to pooling because of personal ID lockers and stuff like that but it would not be a dramatic change and that would need more work anyways. 
- Add a new "NameConfig" field to jobs so we can specify per job naming information
- Pull name datasets out of species and into the job stuff. I mean it's like half pulled out i didnt remove it from species cuz I dont want to touch that yet.
- Lot of dead code removal. how awesome.

## Media

N/A


moonheart08 (approved):
The code looks good but my 2c:
- This may lead to copypaste if we end up with a whole set of jobs using the same modified nameconfig for whatever reason
- are there any jobs we could use this on Now to essentially get a feel for how we'd use this?

EmoGarbage404:
> * This may lead to copypaste if we end up with a whole set of jobs using the same modified nameconfig for whatever reason

Valid and although pulling this out into a prototype isn't hard, I think it would be more burdensome in YAML than necessary for something that (at the moment) is really just meant for a handful of tweaks for specific jobs.

> * are there any jobs we could use this on Now to essentially get a feel for how we'd use this?

Clown should use some custom name datasets I think but I'll leave that as an exercise for someone else because I really don't feel like writing out 300 clown names.

mirrorcult (approved):

PR #1729, opened 2026-04-10
Add clothing closets to maints spawner

PotatOverlord:
## About the PR

title
I initially went for having them be in seperate weights with much lower chances, but I decided to just have them in one big happy family. It might benefit from being in seperate weights but equal weighting if that makes sense.

## Media

image

Changelog

🆑

add: Job wardrobes now have a chance of appearing in maints.


Resources/Prototypes/_ES/Entities/Markers/Spawners/Random/wardrobe.yml
```diff
@@ -28,6 +28,20 @@
...
+      - id: ESWardrobeClownFilled
+      - id: ESWardrobeDetectiveFilled
+      - id: ESWardrobeJanitorFilled
+      - id: ESWardrobeMimeFilled
+      - id: ESWardrobeReporterFilled
+      - id: ESWardrobeSecurityFilled
+      - id: ESWardrobeScienceFilled
+      - id: ESWardrobeDoctorFilled
+      - id: ESWardrobeParamedicFilled
+      - id: ESWardrobeEngineeringFilled
```
Flareguy, line 100: move all of these to their own bespoke table, probably in a different file since these won't have an explicitly associated spawner

Flareguy (changes requested):
the way this PR handles adding them is generally incorrect since we dont want these in dorms i assume. i.e the colored wardrobe spawner still needs to exist as its own thing

EmoGarbage404 (changes requested):
If we have full job wardrobes in maints or w/e they should definitely be fairly rare compared to just colored clothing. At the very least like 50% chance of colored clothes vs 50% chance of job clothing

PotatOverlord:
image

PotatOverlord:
pretty please can this get merged i have to change a single locker spawner every time any map change happens

PR #1730, opened 2026-04-10
Small plasma can

PotatOverlord:
## About the PR

I tested it a bunch and this (about 1/3) hits the sweet spot of 1 swap in a shift to cover the AME while being an amount less. If less plasma in the can is desired then I can tweak the consumption values on the antimatter
Small problem is the can starts off in the yellow light thing but Idk seems like a marginal amount of effort to fix

### DIFF
--- a/Resources/Prototypes/Entities/Structures/Storage/Canisters/gas_canisters.yml
+++ b/Resources/Prototypes/Entities/Structures/Storage/Canisters/gas_canisters.yml
@@ -395,7 +395,7 @@
       gasMixture:
         volume: 1500
         moles:
-          Plasma: 750
+          Plasma: 900
         temperature: 293.15
     - type: Destructible
       thresholds:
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/Entities/Structures/Storage/Canisters/gas_canisters.yml
```diff
@@ -395,7 +395,7 @@
       gasMixture:
         volume: 1500
```
EmoGarbage404: just decrease the volume proportionally as well so the light looks correct

EmoGarbage404 (changes requested):

Resources/Prototypes/Entities/Structures/Storage/Canisters/gas_canisters.yml
PotatOverlord: image

PotatOverlord:
on the timeframe of can swaps, it lasts about 20 minutes at 6 kpa and then the SMESes tide you over for roughly 15 minutes so in theory it's one swap a shift.

PR #1733, opened 2026-04-11
traitor mask guidebook pages

Flareguy:
## About the PR
adds guidebook pages for all of the traitor masks. also updates the existing infiltrator guide to be more informative, in-line with the new guidebooks

## Media
slightly outdated so look at the pr if you want to see the actual contents of it

Content Client_zc43sl7xeM


EmoGarbage404 (changes requested):
Put empty tip sections at the bottom as it'll be hidden for now but automatically show up once we add tips for the corresponding masks

PR #1747, opened 2026-04-14
jerrycan sprite touchup

Flareguy:
## About the PR
makes the jerrycan fit the artstyle more. it was too shiny before and looked unfitting

## Media
Content Client_RAgrpceVyM


mirrorcult (approved):
if you  say so

PR #1753, opened 2026-04-16
custom hud layouting + scaling fixes

mirrorcult:
## About the PR

new control which lays out a leftpanel/viewport/rightpanel hud correctly (lays out viewport first with as much space as possible then puts the panels on either side). evenetually i want to do a splitcontainer type thing and allow you to control the panel ratio by just dragging whcih shouldnt be that hard

scaling now checks 0.5 increments. if it fits better to one of those then itll ditch integer scaling. it should maybe check even more granularly (like 0.25 or something) we'll see i guess. or use a different method for doing this. honestly the previous method probably would have worked fine if we just tweaked it and it basically did exactly this

it should hopefully scale to 720p and 1440p correctly now

this also gets rid of the paneling above/below the viewport because 1) i think it was causing issues and its annoying and 2) i dont like it and its annoying. this makes the layout a bit less stupid to look at sometimes like in stagehand idk

## Media

image
image
image


moonheart08 (approved):

mirrorcult:
decided to check .25 increments also just to make sure

PR #1754, opened 2026-04-16
Radstorm navigation/propulsion equipment

EmoGarbage404:
## About the PR

Adds the navigation terminal and propulsion thrusters. When these devices become unpowered, it'll speed up the radstorm up to 2x.

also rewrites some of the radstorm messages to be warp drive based instead of telesci

## Media

ESToast-0


mirrorcult:
dose the south thruster fit the shuttle still

EmoGarbage404:
> dose the south thruster fit the shuttle still

image

tight squeeze

mirrorcult:
good enough

mirrorcult:
there probably needs to be a little more signposting in maints showing the directions of the thrusters. maybe like directional signs for it

PR #1755, opened 2026-04-16
auto changelog action

mirrorcult:
## About the PR

taken from funky which is taken from dv or w/e

will test and add changelogs for this test in future pr

## Media

N/A

:cl:
Playtest:
- add: In-game changelog support! Probably


mirrorcult:
not sure if it will proc on the same pr adding the action but

PR #1761, opened 2026-04-17
ambient music updates

Flareguy:
## About the PR
adds security and command ambience. also makes atmos play engineering ambience more consistently since it played maint ambience before due to the plating

## Media

N/A


Resources/Prototypes/Entities/Structures/Wallmounts/Signs/posters.yml
```diff
@@ -97,6 +97,7 @@
   components:
   - type: ESPoster
     state: poster2_contraband
+  - type: ESAmbientMusicMarkerAtmos
```
mirrorcult:
```suggestion
  - type: ESAmbientMusicMarkerEngineering
```

PR #1766, opened 2026-04-18
Fueled thrusters

EmoGarbage404:
## About the PR

changium:
- fuel thrusters
- big heater
- mapping changes
- fixes for announcements

## Media

N/A

**Changelog**

:cl:
- add: The station now has several thrusters which keep it moving away from the radstorm. Make sure these are fueled up with welding fuel, or else you may find the storm approaching faster than expected.


Sadie-silly:
engines arent real. What are you doing?

PR #1778, opened 2026-04-19
WizDen space wind fix port

EmoGarbage404:
## About the PR

Ports changes from https://github.com/space-wizards/space-station-14/pull/43471

Kinda undecided on this. It fixes all the false positives on space wind but in return it really kinda under reports. Especially for large spacing, the pressure diff eventually becomes so gradual that you don't see any wind outside of where the breach is. i'm just leaving this up for now until i decide how to feel/tweak it.

## Media

N/A

**Changelog**


ArtisticRoomba:
> It fixes all the false positives on space wind but in return it really kinda under reports.

Partially a consequence of slow spacing/LINDA, when spacing spreads out it has more of an area to pull and as such deltas become smaller and smaller. You can change the epsilons if you want but there's a lot of other hardcoded checks in `AtmosphereSystem.LINDA/HighPressureDelta` that kinda gets in the way of that.

mirrorcult (approved):

PR #1785, opened 2026-04-25
rm spare armor vest from secoff locker

PotatOverlord:
## About the PR

title
Having a spare armor vest means sec can distribute them among sec aligned people, strengthening the both of them considerably as the armor vests are very impactful. I don't actually mind this very much because enlisting crew is pretty cool, but as it is, it scales with the amount of seccies on station when imho it would be nicer as maybe a singular vest that's mapped (I didnt do that in this pr).

Idk if we even want this but I thought I'd pop it up just in case.

**Changelog**

:cl:
- fix: secoffs will no longer have a spare armor vest.


EmoGarbage404:
I dont think i want to do this until we find a more logical way of handling the presence of extra armor on station. i think it's good to have gear that people can break in and steal if they're trying hard enough

PR #1789, opened 2026-04-29
Add a command to simulate mask distribution for masquerades

EmoGarbage404:
## About the PR

Adds the command `mq:sim <masquerade> <playerCount> <trials>` which produces a mask distribution at a given player count with n = `<trials>`

Generally just a nice balancing tool to see how often certain roles show up in a given mode, since our selection tends to be a bit indirect at times.

## Media

image

**Changelog**

none.


moonheart08 (approved):
glad i made masquerade logic idempotent

PR #1793, opened 2026-05-02
allow braincrit people to succumb with an action again

mirrorcult:
## About the PR

being in braincrit already sets mobstate critical, this just readds the actions. no deathgasp because i ddecided i didnt care

kills brain when you succumb

## Media

image

image

image

**Changelog**

:cl:
- add: Being in braincrit now allows you to succumb with an action once more (or say last words)


mirrorcult:
ftr you could always basically succumb by just `/ghost`ing but i think a lot of people didnt realize that and were upset at being made to wait a while for an obvious death

PR #1816, opened 2026-05-05
Increased jerrycan inventory size

### LINKED ISSUE #1772: Jerrycans are too small
should be at least a 2x2 or a 2x3

Coolsurf6:
## About the PR
Jerrycans are now a 3x3 instead of 1x2

fixes #1772

## Media
image

**Changelog**

:cl:
- tweak: Jerrycans are now 3x3


Flareguy:
they should probably be a 2x3 or 2x4 since they hold a lot of reagent in them

Coolsurf6:
> they should probably be a 2x3 or 2x4 since they hold a lot of reagent in them

What about 3x3? Jerry cans being 2x4 would look a bit weird. 2x4 for tanks, 3x3 for containers.

Flareguy:
sure

Coolsurf6:
image

Done, but wanted to bring up that being evil and using shakers now fits more in a 2x3 slot than a 3x3. I think we honestly should curb the shaker and make it a 2x2, I've abused shakers too much for their size way too often.

Flareguy:
i mean if you want to do that that is fine too i think

PR #1827, opened 2026-05-07
New maints utility items: door jammer & net

mirrorcult:
## About the PR

new lovelies

see changelog for how they work

reasoning: lovely items 

actual reasoning: in general, even upstream, we have a lack of interesting utility items that are actually reasonably accessible to people. 

these are both pretty interesting: the door jammer lets you create impromptu locked rooms at a moments notice for many use cases, and the net lets you steal items from people without having to down them first, provided you can actually get by them and pick the item up before they do. also has the utility of letting you delay a gank by taking someones gun or melee away from them, and has some counterplay with the way it functions w active hand

## Media

https://github.com/user-attachments/assets/d34e601d-2b96-4686-920d-4d55c2f07bf1

**Changelog**

:cl:
- add: Added two new common maints utility items: the door jammer & the net. 
- add: The door jammer is a jury-rigged screwdriver that forcibly bolts any door it's thrown and embedded into, until it's removed. 
- add: The net is an item that steals the item out of someone's active hand when it's thrown at them, carrying it along until it lands. If the target's active hand is empty, they'll simply catch the net instead.


Content.Shared/_ES/DoorJammer/ESDoorJammerComponent.cs
```diff
@@ -0,0 +1,16 @@
+namespace Content.Shared._ES.DoorJammer;
+
+/// <summary>
+///     Bolts a door while embedded in it.
+///     Once removed, the door is unbolted, assuming that the door wasn't already bolted.
+/// </summary>
+[RegisterComponent]
+public sealed partial class ESDoorJammerComponent : Component
```
EmoGarbage404: missing networked component for prediction

Content.Server/_ES/Net/ESNetComponent.cs
```diff
@@ -0,0 +1,23 @@
+using System.Numerics;
+using Robust.Shared.Audio;
+
+namespace Content.Server._ES.Net;
+
+/// <summary>
+///     An item which, when thrown, will attempt to yoink an item out of the hands of anyone it hits.
+/// </summary>
+[RegisterComponent]
+public sealed partial class ESNetComponent : Component
```
EmoGarbage404, line 9: i kinda despise that this cant be predicted due to collision stuff but that's the way it goes ig

EmoGarbage404 (dismissed):
Exercise in nonstandard boolean syntax

Content.Server/_ES/Net/ESNetComponent.cs
mirrorcult: it doesnt really make much sense to predict anyway i think

PR #1832, opened 2026-05-08
Update engine

mirrorcult:
## About the PR

wahoo hrtf

**Changelog**

:cl:
- tweak: Positional audio should now sound much better (this is hard to explain)


mirrorcult:
didnt realize this also has the dependency sourcegen stuff in it might need to do something about that

PJB3005:
there's a code fix have fun

mirrorcult:
took a moment but casual +2318 files changed

PR #1841, opened 2026-05-09
bag of fixes

### LINKED ISSUE #1786: Remove CryostasisFactorComponent from mobs
This scales metabolism rate linearly with temperature which, from gameplay testing, plays really fucking bad. If this is necessary for cryopods then it should just be done in a bespoke manner for that. Players getting cold and space and then having their metabolism slow to a crawl fucks up balancing and processing of a lot of shit.

If we want to maintain this behavior, transform the cold/very cold thresholds into status effects and then put discrete levels on there.

### LINKED ISSUE #1804: warp drive charging lies because it accumulates time even when interrupted
needs to offset by lastinterruptiontime

that is why i added it i think i just forgot to do that part

### LINKED ISSUE #1806: increase inhalation threshold
so its only when more visible

### LINKED ISSUE #1807: tcomms processor examine typo
says "preprocessoror" or something like that

### LINKED ISSUE #1810: cuffbreak doafter should be much faster
like 5 seconds or something. go hog wild

### LINKED ISSUE #1840: constructing walls manually needs to take much longer

mirrorcult:
## About the PR

fixes #1807 
fixes #1840 
fixes #1810 
fixes #1806 
fixes #1786 
fixes #1804 

## Media

N/A

**Changelog**

:cl:
- tweak: Breaking out of cuffs, bolas, and welded/locked lockers is now significantly faster in all cases
- fix: Warp drive charge timer no longer ticks up while interrupted
- fix: Being in space no longer makes you take like 9999 minutes to die because of cryostasis
- tweak: Smoke inhalation now has a higher reagent threshold and you'll more reliably cough as a warning at lower thresholds
- tweak: Regular walls & girders now take longer to construct manually (without an RCD) and take less time to deconstruct

### DIFF
--- a/Content.Server/Resist/ResistLockerComponent.cs
+++ b/Content.Server/Resist/ResistLockerComponent.cs
@@ -7,10 +7,10 @@ namespace Content.Server.Resist;
 public sealed partial class ResistLockerComponent : Component
 {
     /// <summary>
-    /// How long will this locker take to kick open, defaults to 2 minutes
+    /// How long will this locker take to kick open
     /// </summary>
     [DataField("resistTime")]
-    public float ResistTime = 120f;
+    public float ResistTime = 10f;
 
     /// <summary>
     /// For quick exit if the player attempts to move while already resisting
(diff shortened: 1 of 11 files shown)


mirrorcult:
might go back on the pull cuff change actually still thinking about it

mirrorcult:
i did

PR #1844, opened 2026-05-09
Denote troupe vs. mask objectives in ui

### LINKED ISSUE #1839: objectives should be clearer about when they're shared (troupe) vs individual
people get confused easily about this. some kind of icon on the objective which can be hovered over for a tooltip saying its a sharead objective would be good (maybe use the troupe icon?)

EmoGarbage404:
## About the PR

adds little headers to the top-right of objective controls to denote whether an objective is a shared troupe objective or a mask objective. Hopefully this will help clarify to people that no, you do not need to do every single traitor objective solo.

had to do mild evil bastard ui code to make the titles stop freaking out but uh it works.

## Media

image

**Changelog**

:cl:
- add: Objectives now state in the UI whether or not they belong to a mask or to a troupe.


mirrorcult:
i still think it's maybe not obvious enough with traitor for example that they're shared (since there's no mask objectives to compare to for them) but

EmoGarbage404:
image
image

updated

mirrorcult:
blessed

PR #1860, opened 2026-05-10
Force round end when all players are dead

### LINKED ISSUE #1742: Round should end when everyone is dead
This is like an implicit behavior of radstorm code which is Fine. but like. just end it if everyone dies? There's not really a point to keep it going lol

mazziebot:
## About the PR
If everyone dies the round ends. Yay!

Closes #1742

## Media

N/A

**Changelog**

:cl:
- tweak: The round now ends automatically when all players are dead.

### DIFF
--- a/Content.Server/_ES/Radstorm/ESRadstormRoundEndRuleSystem.cs
+++ b/Content.Server/_ES/Radstorm/ESRadstormRoundEndRuleSystem.cs
@@ -90,8 +90,6 @@ public sealed partial class ESRadstormRoundEndRuleSystem : GameRuleSystem<ESRads
         {
             component.RadstormNextDamageTickTime = _timing.CurTime + TimeSpan.FromSeconds(1);
 
-            // HELL EVERLASTING! DIE FOREVER!
-            var stillAlive = 0;
             // this should probably not be bounded to mobstate and instead be its own thing but whatever
             var killQuery = EntityQueryEnumerator<MobStateComponent, DamageableComponent, TransformComponent>();
             while (killQuery.MoveNext(out var mob, out var state, out var damageable, out var xform))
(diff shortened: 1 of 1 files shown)


Content.Server/_ES/Radstorm/ESRadstormRoundEndRuleSystem.cs
```diff
@@ -125,6 +125,22 @@ protected override void ActiveTick(EntityUid uid, ESRadstormRoundEndRuleComponen
...
 
+        // If everyone's dead, end the round
+        var mobQuery = EntityQueryEnumerator<MobStateComponent>();
+        var allDead = true;
+        while (mobQuery.MoveNext(out var mob, out _))
+        {
+            // When someone dies they're turned into an observer with no MobState
+            if (HasComp<ActorComponent>(mob))
+                allDead = false;
+        }
```
mirrorcult, line 122: this is imo a mildly brittle way to go about it, id prefer if it at least also checked that the mob's state is actually non-dead before setting `allDead` to false

i think it'd also be better to bound the query on `ActorComponent` instead (theres generally less actors than mobs with mobstate) and check for if it has mobstate + if their mobstate is non-dead

mirrorcult (changes requested):
theres some logic right above this that also handles the case of ending the round when everyone is dead (but only post-radstorm), i think it would make sense to get rid of that logic (i.e. it just damages stuff and doesnt check anything) and just always handle it in the code you've added

mirrorcult:
it would probably also be better to only check this event-based when a player is killed (i.e. sub to `ESPlayerKilledEvent` and check there) though i dont consider it ultra necessary and there are some times ive seen where the kill event seemingly fails to get sent (when theyre dc'd etc) so this is at least guaranteed to catch more cases

Content.Server/_ES/Radstorm/ESRadstormRoundEndRuleSystem.cs
mazziebot: I accidentally committed the wrong version, the correct version is the other way around

PR #1891, opened 2026-05-19
Tragedian

### LINKED ISSUE #1830: Tracking issue for Tragedian mask
https://ephemeralspace.github.io/docs/design/masks/crew/tragedian.html

strict requirements
- objective to embody some concept from a dataset, and the dataset itself with at least a dozen or two good ones (should probably be randomized when the objective is spawned and the logic here just attached to the objective itself). we have an entirely custom objective system, should hopefully be easy to work with, see other masks that spawn with objectives
- action that forces all stagehands to follow the tragedian on a ~long cooldown (5-10min?)
- when the tragedian dies, call a stagehand yes/no vote (we have a custom entity-based vote system, search `ESVote` to see the vote entities and stuff) and mark the objective as completed or failed based on the result of the vote

the mask does not need to be integrated into masquerades, i will handle that

EmoGarbage404:
## About the PR

Implements https://ephemeralspace.github.io/docs/design/masks/crew/tragedian.html

Closes #1830 

## Media

image

image

image

**Changelog**

:cl:
- add: Added the Tragedian, a crew mask who must act out a theme for the approval of stagehands.


mirrorcult (approved):

mirrorcult:
lint fail real

PR #1892, opened 2026-05-23
remove throwing damage from cans

PotatOverlord:
## About the PR

title
it really really sucked because people would throw them willy nilly on the shuttle and cause like 50 damage per throw since everyone was bunched up
left it on glass and gold because they break as soon as they're thrown
possibly could just make it do like 1 damage on throw so its still something 
the whole drinks section needs a yml overhaul because the parenting sucks 

**Changelog**

:cl:
- fix: cans no longer deal damage upon being thrown


piskaczek:
you really did do it huh

piskaczek:
valid creative writing outlet

PR #1912, opened 2026-05-24
Arsonist tweaks

### LINKED ISSUE #1876: Arsonist should get objective progress for their fires spreading
The arsonist objective only counts the fire that they directly create (spark ignitions, firebombs, welding tank explosions, etc.) but those fires can continue to spread and get huge, which is not counted at all.

EmoGarbage404:
## About the PR

Closes #1876 

changes:
- objective now ticks for fire spreading
- increased objective limit

arsonist's objective completion is not even remotely proportional to the amount of damage they cause which is stupid. whole damn station on fire and they dont get a win for it what the FUCK.

## Media

N/A

**Changelog**

:cl:
- tweak: Arsonists' objective now progresses when their fire spreads. To compensate, they need to create more fires.


CheeseWizard2:
If fires smolder out now is it possible that that could decrement arsonist points?

CheeseWizard2:
to clarify i think that's something that should be prevented coz i think it'd be mega annoying

PR #1916, opened 2026-05-25
APC Breakage

EmoGarbage404:
## About the PR

Adds a generic breakage system for more easily making destruction/breakage behavior.

revamps APC to not use construction and to instead have a flat breakage state.

## Media

image

**Changelog**

:cl:
- add: APCs now become broken when damaged. In this state, they do not provide power.
- tweak: APCs can no longer be constructed.


Resources/Prototypes/Entities/Structures/Power/apc.yml
```diff
@@ -124,29 +137,18 @@
...
       behaviors:
-      - !type:SpawnEntitiesBehavior
-        spawn:
-          SheetSteel1:
-            min: 1
-            max: 1
       - !type:DoActsBehavior
-        acts: [ "Destruction" ]
+        acts: [ "Breakage" ]
       - !type:PlaySoundBehavior
```
mirrorcult: does it really make sense to have this still? since we're kind of just sidestepping destructible entirely

mirrorcult (approved):

PR #1919, opened 2026-05-28
Correct popup prediction

EmoGarbage404:
## About the PR

Popups are plagued with a ton of confusing overload methods designed to handle cases when prediction causes duplicate popups to be shown. This leads to what can only be described as an incredibly confusing architecture where each caller has to individually consider whether they are currently operating in a predicted environment (impossible to know). As such, it's fairly common to see double popups or have missing popups in situations where the incorrect method is getting called in the wrong environment.

This PR corrects this by adding proper prediction collision resolution to popups. Using a similar system as Pain Flashes, the client now creates tokens for each popup, which it uses to resolve collisions by not showing popups coming from the server with matching tokens.

Tested to the best of my ability (there were like 500 popup usages i had to redo) so if you want to run around in this beautiful world let me know if anything is broken.

## Media

**Changelog**

### DIFF
--- a/Content.Client/Popups/PopupSystem.cs
+++ b/Content.Client/Popups/PopupSystem.cs
@@ -1,5 +1,4 @@
 using System.Linq;
-using Content.Shared.Containers;
 using Content.Shared.Examine;
 using Content.Shared.GameTicking;
 using Content.Shared.Popups;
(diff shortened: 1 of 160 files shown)


EmoGarbage404:
i hope this doesnt break anything

Content.Client/Popups/PopupSystem.cs
```diff
@@ -108,6 +109,10 @@ private void PopupMessage(string? message, PopupType type, EntityCoordinates coo
         #region Abstract Method Implementations
         public override void PopupCoordinates(string? message, EntityCoordinates coordinates, PopupType type = PopupType.Small)
         {
+            if (!Timing.IsFirstTimePredicted || message is null)
+                return;
+
+            _predictionInstances.Add(new PopupCoordinatesEvent.PredictionInstance(message, type, Timing.CurTick, GetNetCoordinates(coordinates)));
```
slarticodefast, line 115: Pretty sure this did not work when I tested this if you have something like the player ashing and you show a popop at their location they had when they got deleted (which is the main reason to use PopupCursor over PopupEntity) due to predicted movement. The coordinates will differ between server and client, so it gets shown twice if the player is moving.

PR #1920, opened 2026-05-30
remove detvest and cap forensic scanner

Flareguy:
## About the PR
removes the detvest. i dont like that det can wear at no downside and it looks worse then the coat + it makes det more distinct since they already get the gun and the forensic scanner.

also removes the captains forensic scanner for similar reasons

## Media

N/A

**Changelog**

:cl:
- remove: The detective no longer starts with an armor vest.
- remove: The captain no longer starts with a forensic scanner.


Sadie-silly:
Yes! This is pretty good.

Ermucat:
Yes! This is pretty good.

PR #1941, opened 2026-05-31
Fix caches playing double popups

### LINKED ISSUE #1935: Cache drop message mispredicts
popup should just be on the entity

EmoGarbage404:
## About the PR

Closes #1935 

the predicted spawn gives two unique entityuids which means trying to do a popup on that will register as trying to do two unique popups. fml.

## Media

N/A

**Changelog**


mirrorcult (approved):

mirrorcult:
should probably have a good way to deal with that later but uhh

PR #1968, opened 2026-06-13
remove spare IDs from hops vendor

PotatOverlord:
## About the PR

title
you could just get a ton to overwrite and have AA again

## Media

N/A

**Changelog**

no cl


EmoGarbage404:
I don't really think this is necessary since I already greatly reduced the number of spare IDs HOP has access to. I think HOP's ability to set access is relatively worthless if they have no extra IDs for actually doing so

PotatOverlord:
Thats fair yeah should've probably waited to see how it plays out as well

PR #1995, opened 2026-06-14
More rare maintenance loot

### LINKED ISSUE #1966: Traitor maints gear is too rare
more

CrystalineSpider:
## About the PR
Resolves #1966

Increased the chance of "rare" maintenance loot, including traitor items.
They're around 4 times more common, and I saw around 4-10 in a wide sweep of maintenance.

The chance for pistols from maintenance also reduced to 1/3 previous.

Personally I feel this is a bit on the lower side given what is actually in the rare pool, but I'll leave it and come back to it if we feel like it.

## Media

N/A

**Changelog**

:cl:
- tweak: Increased the odds of traitor, and likewise rare items, from maintenance

### DIFF
--- a/Resources/Prototypes/_ES/Catalog/Fills/Crates/crates.yml
+++ b/Resources/Prototypes/_ES/Catalog/Fills/Crates/crates.yml
@@ -413,9 +413,10 @@
     - id: ESSleepyPen
       weight: 0.5
     - id: C4
-    - id: ESWeaponRevolver357
-    - id: ESWeaponPistol9mm
-    - id: ESWeaponPistol9mmSilenced
+    - group:
+      - id: ESWeaponRevolver357
+      - id: ESWeaponPistol9mm
+      - id: ESWeaponPistol9mmSilenced
     - id: ESClothingOuterArmorSecurity
     - id: ExGrenade
     - id: ESSpeedLoader357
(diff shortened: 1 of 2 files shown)


mirrorcult:
this is a little bit too much of an increase for me to be comfortable wit tbh

mirrorcult:
think like weight 3-4 on rare would be fine

Resources/Prototypes/_ES/Catalog/Fills/Crates/crates.yml
```diff
@@ -413,12 +413,14 @@
...
       weight: 0.5
     - id: C4
-    - id: ESWeaponRevolver357
-    - id: ESWeaponPistol9mm
-    - id: ESWeaponPistol9mmSilenced
+    - group:
+      - id: ESWeaponRevolver357
+      - id: ESWeaponPistol9mm
+      - id: ESWeaponPistol9mmSilenced
+      weight: 2
```
mirrorcult: should be regular weight imo or even lower weight than usual

CrystalineSpider: This was done mostly as with the current weight they were simply too common. Likely this will change based on what I see spawning, personally I think normal weight will be pretty likely once it's toned down.

CrystalineSpider:
Effectively halved spawns, tbh it doesn't feel that much different then the current just because items like pinpointers aren't especially flashy in localhost. Better than the 1% chance it had per roll before though for sure.

mirrorcult:
by normal weight for the guns i meant still grouped just with weight 1 instead of weight 2

CrystalineSpider:
> by normal weight for the guns i meant still grouped just with weight 1 instead of weight 2

Ah

PR #2028, opened 2026-06-20
minor mapping fixes for 6/20 playtest

Flareguy:
## About the PR
adds firebots and medibot to shuttle
also moves lathe inwards so its more defendable

## Media

N/A

**Changelog**

:cl:
- tweak: The cargo lathe is now inside of cargo.
- tweak: There are now helpful bots on the shuttle.

### DIFF
--- a/Resources/Maps/_ES/Shuttles/arrivals.yml
+++ b/Resources/Maps/_ES/Shuttles/arrivals.yml
@@ -1,11 +1,11 @@
 meta:
   format: 7
   category: Grid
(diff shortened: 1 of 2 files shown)


mirrorcult:
not really sure i agree with moving the lathe into the dept but we'll see

mirrorcult (approved):

mirrorcult:
also dont really agree with this solution for shuttle stuff i think people are probably just going to drag them off the shuttle

PR #2029, opened 2026-06-20
nerf revolver

Flareguy:
## About the PR
changes revolver damage from 30 to 40. revolver felt too strong so this makes it less so hopefully. this puts it from 3 shots to crit to 4
it still retains its armor piercing properties even on opponents with the max 5DT, meaning its basically just the same but with 1 more shot required

## Media

N/A

**Changelog**

:cl:
- tweak: Nerfed revolver damage slightly. It now takes 4 shots to down instead of 3.


Sadie-silly:
Do you even play the game? You're ruining the meta for no reason. These devs never listen to their players .

mirrorcult:
id like the firerate to go down a tad on top of this i think

mirrorcult:
ehh whatev

PR #2030, opened 2026-06-20
zero g tweaks experiment

mirrorcult:
## About the PR

testing stuff out

also reworks damageonhighspeedimpact to actually function at all

## Media

https://github.com/user-attachments/assets/f96e9793-7ca8-4900-aa80-24ba0c452998

**Changelog**

:cl:
- add: Zero G movement has been changed substantially: you must now be near walls even when on station to move effectively, acceleration is much higher (and max speed is muuuuch higher, you can go quite fast), and its easier to stop with no inputs held when near walls
- tweak: Gravgen now runs out of power & charges faster


mirrorcult:
also getting rid of the old playtest changelog thing in here

PR #2053, opened 2026-06-22
damaged lattices & tiles

mirrorcult:
## About the PR

flareguy sprites

had to change a lot of stuff to properly use tile variants so thats nice for other things as well

smoothing some day so we can have actually decent looking lattices

anyway motive here is to retain the 'structure' of an underlying station even if it gets heavily damaged by destruction in the same vein as i want damaged walls and machine repair and stuff

## Media

**Changelog**

:cl: Flareguy, mirrorcult
- add: Explosions/meteors can no longer reduce a tile to space, instead it is replaced by damaged lattice which can be repaired with a metal rod back to a normal lattice
- add: Explosions will now spawn saucy damage overlays on most tiles they damage


mirrorcult:
doing minor form of tile damaging in this as well should be easy

Sadie-silly:
its gonna be so peak

mirrorcult:
done

PR #2082, opened 2026-06-29
KILL THE WORLD

mirrorcult:
## About the PR

prune rsis not referenced in any yaml or c# client startup time go up

scripot i used https://gist.github.com/mirrorcult/a21e1a36cb2908ec3012f9069e951159 not putting in repo cuz i was doing it with ripgrep initially so i wrote the script to just invoke ripgrep and also i was running it from project root and IDK its not like we'll have to do this that many more times probably... i guess once we clear out mroe shit but idk (the script doesnt remove anything it just outputs a list of unused rsi paths in a file)

image


mirrorcult:
you tried Oldbuddy

PR #2121, opened 2026-07-04
Warp drive small fixes

### LINKED ISSUE #2106: Warp drive clogs too often
Either it's bugged or the way it's done is insane. It's clogging every few minutes and just constantly needing to be cleared out. the time between clogs is also really low.

EmoGarbage404:
## About the PR

Closes #2106 

fixes an incorrect check for ghosts

delays the next clog time set code until after the warp drive is unclogged. this should mean that you always have _at least_ 5 minutes after unclogging until it happens again. Before, the time would be set when the clog occurs, meaning that if it takes you 4 minutes to unclog it, you'd get another clog 1 minute after. That feels insanely bullshit and punishing so this will hopefully feel better.

## Media

N/A

**Changelog**


mirrorcult:
i might be dumb

PR #2142, opened 2026-07-06
fix viewport layouting for good

mirrorcult:
## About the PR

this is ~~not~~ done theres still a bunch of random bugs i have~~nt~~ figured out

yaaaaay! fixes most bugs i could find wrt resizing and wrong snap and shit

:cl:
- fix: Viewport stuff should now behave much more sanely with resizing
- fix: Viewport stuff should no longer crash when the chat ends up resized too small from word wrap stuff
- tweak: Viewport container left/right panels now go invisible if not given enough space instead of being like 5px wide
- tweak: Viewport scale now snaps to integer sizes more aggressively (and more sanely) in cases where it looks fine


mirrorcult:
ready

image

PR #2146, opened 2026-07-06
post-death sequence

mirrorcult:
## About the PR

nice to see the immediate aftermath of your death. esp if you're a worm or something, while still keeping the -> lobby transition. and its aura of course. curtains city

## Media

https://github.com/user-attachments/assets/f45f1414-7835-4163-a33a-3152176ac5f6

**Changelog**

:cl: mirrorcult,SadieFei
- add: Rather than instantly being transported to the lobby, a short little post-death sequence will play out when you die
- tweak: Use of the ghost command is now only allowed if you are dead (and want to expedite the death cutscene for some reason)


mirrorcult:
this wont work properly for burstworm since they get gibbed not sure how i should handle it tbh

Sadie-silly:
what if instead of just like liquefying the corpse when they burst it turns your corpse into like a mangled unrecognizable mess? Kind of like how the broken lattice tile replaces tiles getting just turned into space so we don't have to deal with any weird jank that comes from tile destruction

EmoGarbage404:
> this wont work properly for burstworm since they get gibbed not sure how i should handle it tbh

spawn a temporary "eye" entity at the position of death that you just attach the player too. You can either always do it (if you dont want the alerts visible) or only do it when the body is gibbed.

mirrorcult:
ill probably just always do it to avoid any potential weirdness

mirrorcult:
image

fixed the gibbing case forgot to video it but it did work + aftermath

PR #2148, opened 2026-07-07
cinematics + nuke cinematic

mirrorcult:
## About the PR

draft mostly finished tho. some unanswered questions with how i want this to interact with roundend + how best to make things like chat not work while the cinematic is happening


mirrorcult:
fixed up and works for the nuke cinematic. no video you'll see it ingame

mirrorcult:
radstorm and warpdrive ending cinematics later idk i dont feel like speedrunning that rn

PR #2194, opened 2026-07-15
mirror mode variation

mirrorcult:
## About the PR

10% (will probably lower later once it gets seen) chance for the map to spawn completely flipped

surprising amount of effort to fix things to work with this, interaction particles were brokey by accident with camera rotation which i think was my fault

had to spritecomp `noRot` items since its something we want anyway and the only real blocker on it was the throwing animation which i converted to use an actual sprite animation instead of physics impulse

## Media

https://github.com/user-attachments/assets/05495579-e02d-4e9e-a38d-be1bf3ac7e09

**Changelog**

:cl:
- add: A very very very very very scary flipped mirrored reality may enter this world with a 10% chance.
- tweak: Camera rotation binds no longer rotate the camera


mirrorcult:
image

Sadie-silly:
oh it's peak

PR #2199, opened 2026-07-16
make stagehand vote container invis if there are no vounts

mirrorcult:
## About the PR

ya

## Media

image
image

**Changelog**

:cl:
- tweak: Stagehand vote container will now be invisible if there are no active votes


mirrorcult:
vounts

PR #2200, opened 2026-07-16
reverb actual

mirrorcult:
## About the PR

requires engine pr

## Media

N/A

**Changelog**

:cl:
- add: A base level of reverb has been applied to positional sounds, will be more dynamic based on room size in the future


Sadie-silly:
exciting!

PR #2237, opened 2026-07-18
barricade kit and secret wall kit sprite updates

Flareguy:
## About the PR
updates barricade kit and secretwall sprites slightly
no media because its a minor change just look at the rsi diff

## Media

N/A


mirrorcult (approved):
if you say so

PR #2238, opened 2026-07-18
Filth / Miasma

EmoGarbage404:
## About the PR

Adds miasma, miasma atmos effects, miasma clouds, etc. etc. all sorts of goodies.

doesn't implement diseases rn because i cant be fucked

based on this thing:
https://ephemeralspace.github.io/docs/design/decay/filth.html

## Media

N/A

**Changelog**

:cl:
- tweak: Bodies now produce miasma immediately after dying.
- add: Blood puddles now produce small amounts of miasma.
- add: Miasma buildup now creates dirt decals and pests.
- add: Miasma buildup can now create miasma clouds, which can be protected against with a bio suit.


mirrorcult:
the miasma arrives

mirrorcult (approved):
miasma shouldnt get scrubbed by d efautl imo

also not sure on the balance of the miasma production but we'll see

PR #2278, opened 2026-07-19
Fix rice grinding

cashew-42:
## About the PR
uhhh i fixed the rice bug, it's also present in upstream so i'll pr it there aswell.

It currently isn't possible to insert rice bushels into grinders, this pr fixes that

## Media

N/A

**Changelog**

N/A


cashew-42:
nevermind I lied, it works in upstream. Upstream has an updated ProduceBase

PR #2286, opened 2026-07-22
marauder changes round 2

Flareguy:
## About the PR
the marauder kit no longer has the esword. they have their revolver and again ammo but instead of the grenade and knife they get 2 door jammers. hopefully this makes them more interesting to play like the traitor other masks

## Media

N/A

**Changelog**

:cl:
- tweak: The marauder kit has been reverted to the previous version. They now also get two door jammers instead of a knife and grenade, now.


EmoGarbage404 (changes requested):
Said this in discord but i kinda want to go back to the old kit except swapping maybe the knife and some of the armor for some kind of interesting unique item. Something like a single use stim or just a general itme thatd help with going on offensives.

PR #2301, opened 2026-07-25
Cryogas & Cryohusks

EmoGarbage404:
## About the PR

Implements cryogas and cryohusks

omits some small aspects of cryohusks (transformed cryohusks don't lose clothing or get random accesses on IDs) ((i think this stuff is kinda unnecessary and would just make them more confusing))

Frezon is now cryogas, which is mostly the same except with a more sane cooling effect and a better sprite. exposure to cryogas when incapacitated has a chance to cryohusk you. Cryohusking is essentially a polymorph that strips all your identifying features. It also turns your Id card into a nice opaque block. 

secondary fix: swapping jumpsuits no longer drops PDAs on the ground

Singular primary issue: damage transfer does not work at all and thus getting cryohusked is a full heal. not terribly far off from what we wanted but i can't be fucked to figure this one out at 2 am

## Media

image
image

**Changelog**

:cl:
- add: Added Cryogas, a chilly gas that can convert exposed bodies into Cryohusks.
- fix: Swapping jumpsuits and outerclothing will no longer drop PDAs, pocket items, or suit storage tanks onto the ground.
- remove: Removed Frezon.


mirrorcult:
gas sprite is so good man

mirrorcult (approved):

PR #2303, opened 2026-07-25
make warp disciple teleport effect work so they dont just randomly dissapear if you have them cuffed or something

FyeStance:


mirrorcult:
mildly confused by the logic here but ill clean it up later since i wanted to tweak something about it anyway

FyeStance:
Me Too

PR #2351, opened 2026-07-29
turns monkeys in to a species & gives them displacement maps

Flareguy:
## About the PR
adds monkeys as a species and gives them displacement maps. useful for ExperimentCoin. doesnt remove old prototypes or do anything to old monkey appearances since i dont want to do that in this pr

also fixes newspaper inhands because i didnt actually make the right hand version different apparently

## Media
Content Client_ZEYJA7AoQe


Resources/Prototypes/_ES/Entities/Mobs/Species/monkey.yml
```diff
@@ -0,0 +1,141 @@
...
+    drawdepth: Mobs
+    layers:
+    - map: ["enum.DamageStateVisualLayers.Base"]
+      state: monkey
+      sprite: Mobs/Animals/monkey.rsi
+    - map: [ "enum.HumanoidVisualLayers.Handcuffs" ]
+      color: "#ffffff"
+      sprite: Objects/Misc/handcuffs.rsi
+      state: body-overlay-2
+      visible: false
```
EmoGarbage404, line 97: i think you need to re-specify the layers in basespeciesorganic for clothes. you mentioned the clothing not being layered correctly and i'm 99% sure this is why

PR #2354, opened 2026-07-29
qm forcegun/tethergun

Flareguy:
## About the PR
the qm now gets a forcegun Or tethergun. im not sure which one is more fun or reasonable so they have a chance to get either in their locker

readds forcegun/tethergun to facilitate this (obviously)

## Media

N/A

**Changelog**

:cl:
- add: The QM can now start with either a tethergun or forcegun.


mirrorcult:
This PR approved by the Fuck It Why Not Coalition

PR #2400, opened 2026-08-04
Microwave is no longer a weapon

### LINKED ISSUE #2384: microwaves are cracked
I pain crit a guy in 3 hits with a microwave what the fuck is this shit

CrystalineSpider:
## About the PR
Resolves #2384

I really did consider for a long time making it instead just a more balanced weapon, but the kitchen already has all the knives and taking the microwave to use (and likely break it) screws over the chef if there is one.

It's funny and I think it feels bad to do but genuinely I see so many reasons to just remove it so I have.

## Media

N/A

**Changelog**

:cl:
- tweak: Microwaves are no longer a melee weapon


EmoGarbage404 (changes requested):
could you also make it not an item

CrystalineSpider:
idk why I didn't just do that

Resources/Prototypes/Entities/Structures/Machines/microwave.yml
EmoGarbage404: pretty sure this is all related to either damage stuff or item/throwing stuff

CrystalineSpider: testing to make sure nothing breaks otherwise fixed, thank you for catching it

PR #2408, opened 2026-08-05
minor combat balancing

mirrorcult:
## About the PR

didnt like how good armor was still

didnt like how revolver is still insanely busted

pistol fired a little too fast i think

## Media

N/A

**Changelog**

:cl:
- tweak: Revolver and pistol shots do a little less damage now
- tweak: Pistols have less max firerate
- tweak: Outer armor has 1 less DT

### DIFF
--- a/Resources/Prototypes/_ES/Entities/Objects/Weapons/Guns/revolver.yml
+++ b/Resources/Prototypes/_ES/Entities/Objects/Weapons/Guns/revolver.yml
@@ -27,7 +27,7 @@
   - type: AmmoCounter
   - type: Gun
     selectedMode: SemiAuto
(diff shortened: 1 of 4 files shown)


Resources/Prototypes/_ES/Entities/Objects/Weapons/Guns/revolver.yml
```diff
@@ -27,7 +27,7 @@
   - type: AmmoCounter
   - type: Gun
     selectedMode: SemiAuto
-    fireRate: 1.5
+    fireRate: 0.65
```
mirrorcult, line 30: this is not a real change revolver already fired this slow, just changing the base to match

PR #2474, opened 2026-08-10
cooldown circle around mouse

mirrorcult:
## About the PR

atm this only renders if you have the show hand item overlay enabled but apparently emo has that turned off so maybe people turn it off??? i wasnt aware people turned it off. do people that turn it off also want this turned off. lets find out(?)

## Media

image

**Changelog**

:cl:
- add: Use delay cooldown circle (like extinguishers or tools or whatever) now shows around the cursor also. assuming you have the 'show hand item' option enabled

### DIFF
--- a/Content.Client/Cooldown/CooldownGraphic.cs
+++ b/Content.Client/Cooldown/CooldownGraphic.cs
@@ -29,7 +29,6 @@ namespace Content.Client.Cooldown
 
         protected override void Draw(DrawingHandleScreen handle)
         {
-            Span<float> x = new float[10];
             Color color;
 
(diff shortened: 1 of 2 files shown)


Content.Client/Cooldown/CooldownGraphic.cs
```diff
@@ -29,7 +29,6 @@ public CooldownGraphic()
 
         protected override void Draw(DrawingHandleScreen handle)
         {
-            Span<float> x = new float[10];
```
mirrorcult, line 32: idk either ok

PR #2478, opened 2026-08-10
Movable selections for mapping

sowelipililimute:
## About the PR

aghosts now can select areas and move their tiles/entities and shit
oh shit i forgot decals uhhhh that's a part 2 honestly especially because we're getting decal chents soon i think
there's a lot that could be done but i'm not sending in a 50kloc PR

## Media
https://github.com/user-attachments/assets/5c8d9f45-0cea-4c28-9e1c-c7fdda767702


mirrorcult:
(Eats pr))

PR #2492, opened 2026-08-13
remove `IsInDetailsRange`

mirrorcult:
## About the PR

eminently pointless only serves to confuse people about why they cant examine something they could before since its not telegraphed at all. if we want conditional examination it should be done differently

merge after chatfactor


Coolsurf6:
God I want to remove this upstream as well it's horrific with bad ping.

PR #2503, opened 2026-08-14
Name tag peeking

jessicamaybe:
## About the PR

hold shift to peek at peoples names

doesn't work if the person your looking at is in the dark

uses the wonderful wormtown9k my beloved

## Media

https://github.com/user-attachments/assets/16b6c974-5a35-4a0d-9b63-f94dc83b0717

N/A

**Changelog**

:cl:
- add: You can now hold shift to peek at player names

### DIFF
--- a/Content.Client/_ES/NamePeek/ESNamePeekOverlay.cs
+++ b/Content.Client/_ES/NamePeek/ESNamePeekOverlay.cs
@@ -1,5 +1,6 @@
 using System.Numerics;
 using Content.Client.Examine;
+using Content.Shared._ES.Viewcone;
 using Content.Shared.IdentityManagement;
 using Content.Shared.Interaction;
 using Content.Shared.Mobs;
@@ -33,13 +34,14 @@ public sealed partial class NamePeekOverlay : Overlay
     [Dependency] private IUserInterfaceManager _uiManager = default!;
     [Dependency] private IPrototypeManager _prototypeManager = default!;
 
-    private readonly EntityLookupSystem _lookup;
-    private readonly LightLevelSystem _lightLevel;
-    private readonly ESNamePeekSystem _namePeekSystem;
     private readonly ExamineSystem _examineSystem;
-    private readonly SpriteSystem _sprite;
-    private readonly SharedTransformSystem _transform;
+    private readonly EntityLookupSystem _lookup;
+    private readonly ESNamePeekSystem _namePeekSystem;
+    private readonly ESViewconeAngleSystem _viewCone;
+    private readonly LightLevelSystem _lightLevel;
     private readonly ShaderInstance _shader;
+    private readonly SharedTransformSystem _transform;
+    private readonly SpriteSystem _sprite;
 
     private EntityQuery<SpriteComponent> _spriteQuery;
     private EntityQuery<TransformComponent> _transformQuery;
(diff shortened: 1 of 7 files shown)


mirrorcult:
blessed

Content.Client/_ES/NamePeek/ESNamePeekOverlay.cs
```diff
@@ -0,0 +1,192 @@
...
+    private readonly LightLevelSystem _lightLevel;
+    private readonly ShaderInstance _shader;
+    private readonly SharedTransformSystem _transform;
+    private readonly SpriteSystem _sprite;
+
+    private EntityQuery<SpriteComponent> _spriteQuery;
+    private EntityQuery<TransformComponent> _transformQuery;
+    private EntityQuery<MobStateComponent> _mobStateQuery;
+
+    private TextOutline _outline = new (2.5f, Color.Black);
```
mirrorcult: outline should be 2px since iirc thats whats used for chat and popups atm

Content.Client/_ES/NamePeek/ESNamePeekOverlay.cs
```diff
@@ -0,0 +1,192 @@
...
+        foreach (var ent in _nearbyEntities)
+        {
+            if (ent.Owner == playerEnt)
+                continue;
+
+            if (!_transformQuery.TryComp(ent, out var xform))
+                continue;
+
+            if (!_viewCone.InViewcone(playerEnt, ent))
+                continue;
```
mirrorcult: imo rather than adding a new keybind and checking for viewcone i think it should probably just be tied to the same key as hold-to-face is (so that you're always going to be looking in the direction of the thing you're examining anyway)

Content.Client/_ES/NamePeek/ESNamePeekOverlay.cs
```diff
@@ -0,0 +1,192 @@
...
+            if (eye.DrawLight && eye.DrawFov)
+                _lightLevel.TryCalculateLightLevel(mapPos, out lightLevel);
+
+            //Don't show nametag if it's too dark
+            //Most maintenance tunnels on toast seem to be below 0.9, main halls are usually 1
+            if (lightLevel < 0.89)
+                continue;
+
+            if (!_spriteQuery.TryComp(ent, out var sprite))
+                continue;
```
mirrorcult: check sprite visibility + alpha also (set alpha of the text to the same alpha as the sprite maybe?)

mirrorcult (changes requested):

mirrorcult:
i really dislike having to check for lightlevel etc, i wasnt even aware of that system (i guess the illumination measuring got into engine) but it looks incredibly inefficient, going to see if there's a way to just have the overlay draw in worldspace sanely

jessicamaybe:
> i really dislike having to check for lightlevel etc, i wasnt even aware of that system (i guess the illumination measuring got into engine) but it looks incredibly inefficient, going to see if there's a way to just have the overlay draw in worldspace sanely

WorldSpace doesn't have DrawString yet so idk what else to do about low light level. 

It might not be that big of a deal because you could just spam right click anyways to see if someones there but idk! feels wrong to be able to magically tell if there is a demon man in the dark

mirrorcult:
afaict glyphs support being drawn with `DrawingHandleWorld` just fine, ill see if i can pull it out or something

mirrorcult:
going to just merge now and issue it + tweak it after

PR #2512, opened 2026-08-15
Radio clicks

sowelipililimute:
## About the PR
Radios now make a little satisfying click sound when you chat on them.

## Media
https://github.com/user-attachments/assets/f132225c-76c8-4dcc-ae30-bd82907dde29

probably could use some audio balancing but i'm not the best person to ask to do that xd

**Changelog**

:cl:
- add: Radios click when you transmit a message.


mirrorcult (approved):
going to make quieter in diff pr

PR #2528, opened 2026-08-16
Remove Stamina

### LINKED ISSUE #2430: Remove stamina
Blackjack test went well and i dont want it cluttering my glorious HUD anymore

EmoGarbage404:
## About the PR

Closes #2430 

The short reasoning is that stamina only makes sense in an environment where you want a secondary combat layer on top of normal combat to allow select people to win fights without necessarily having to engage in combat. in a more equitable power system, this is kinda insane horseshit so... GOODBYE! It's gone.

added stamcrit behavior to stuns so that we could keep that for paincuffing yip yip yippee

## Media

N/A

**Changelog**

no cl cuz it's not really something that is playerfacing anymore


mirrorcult:
im a little wary of removing it since i think the base of "number that goes down with certain actions, regenerates over time, and if it hits 0 you get stunned momentarily" is like. pretty much fine but also i dont really car about any of the uses of stamina atm and it'd require a full rework of it anyway

PR #2532, opened 2026-08-17
Remove emojis from sanitized chat channels

EmoGarbage404:
## About the PR

the regex may be overly aggressive but it gets the emojis and that's all that matters for english AFAIK. doesn't apply on stagehand or lobby chat cuz why should it.

## Media

N/A

**Changelog**

:cl:
- tweak: Emojis can no longer be sent in in-character chat messages.


Sadie-silly:
⚠️🪱

Coolsurf6:
🪱
👗

PR #2557, opened 2026-08-19
Breakable Computers

EmoGarbage404:
## About the PR

updates computers with breakable states, removes the construction associated with them, etc etc you know the drill.

broken computers have a BSOD and also spark occasionally (fire hazard anyone?)

## Media

image

**Changelog**

:cl:
- tweak: Computers now have a broken state and cannot be destroyed fully or deconstructed.


Coolsurf6:
Does this include sabotagable computers like comms or even nuke terminals etc? I'd see them maybe not as people damaging nuke terminals just to maybe stop traitors is kinda odd. 

Also if they do occasionally spark, does that actually cause fires randomly? Huge if true.

EmoGarbage404:
Terminals are unchanged

Content.Server/_ES/Breakable/ESSparkWhileBrokenSystem.cs
```diff
@@ -0,0 +1,29 @@
...
+    [Dependency] private ESBreakableSystem _breakable = default!;
+    [Dependency] private ESSparksSystem _sparks = default!;
+
+    public override void Update(float frameTime)
+    {
+        base.Update(frameTime);
+
+        foreach (var ent in EntityQueryEnumerator<ESSparkWhileBrokenComponent>())
+        {
+            if (!_random.Prob(ent.Comp.SparkChancePerSecond * frameTime))
```
mirrorcult: i think itd be good for this be a component which gets added when a thing is broken rather than always being there and checking for breakable

EmoGarbage404: Opted to just use a boolean on the comp for "enabled" and then check that to avoid spamming out this random call every tick.

PR #2559, opened 2026-08-20
Duffel Caches

### LINKED ISSUE #2394: rework caches
way too easy / noobtrappy to get caught out getting a cache rn even if ur not traitor

instead of spawning in a crate, spawn in a smugglers satchel with all the items (cant have items put into it or something maybe) and put it in your hand when you get the cache. easier to walk away immediately after getting it

EmoGarbage404:
## About the PR

Closes #2394

Caches now come in duffel bags instead of crates. When they are pulled down, they go immediately in your hand which makes them easier to grab while in darkness all stealthy like.

cache tweaks:
- chameleon clothes now come in the cache itself instead of another backpack. duh
- subverter guns have been changed slightly. this is really just because you can't fit 2 shotguns alongside everything else. plus idk maybe this will make them more interesting if its a little bit more consistent.

## Media

N/A

**Changelog**

:cl:
- tweak: Equipment caches now come in the form of duffel bags instead of crates.


mirrorcult:
does the icon have a crate on it, I genuinely don't remember

EmoGarbage404:
> does the icon have a crate on it, I genuinely don't remember

it might

mirrorcult (approved):

PR #2569, opened 2026-08-22
flimsy anti wire underwall protection

### LINKED ISSUE #2127: Explosions break wires covered by walls
for APCs, if an explosion hits the apc, it'll break the wires underneath the wall because they're technically exposed (plating)

PotatOverlord:
## About the PR

Cableplacer will now check if anchored entities on the same tile have the wall collision layer, and if so prevent placement. This doesn't stop people placing them under girders then fixing the wall, but this is mostly a QoL change in my mind.
(my vibe detector is saying its a weird way to do it but it works so idk up to you)

I'm gonna check toast for any underwall wires that are still hanging around but I'm like 70% sure theres none left after the apc pass. Update: nothing except for the random spawners idk idgaf about that
Also stops the RCD from deconstructing wires because like. whys that a thing its just annoying when im trying to deconstruct a window and it hits a wire underneath 4no rasin

half fixes #2127 kinda because theres no more wires under walls

## Media

https://github.com/user-attachments/assets/73991749-3632-4d71-8e6e-2efa64ca6e7c
my preview of this wasnt working idk if its actually showing anything
upddate: it isnt but if you click it it works. idk man

**Changelog**

:cl:
- tweak: you can no longer place wires underneath walls.


Content.Server/Power/EntitySystems/CableSystem.Placer.cs
```diff
@@ -45,6 +48,9 @@ private void OnCablePlacerAfterInteract(Entity<CablePlacerComponent> placer, ref
             if (_whitelistSystem.IsWhitelistPass(component.Blacklist, anchored))
                 return;
 
+            if (_physics.GetHardCollision(anchored).Layer == (int)placer.Comp.BlockingLayer) // To stop placing wires under walls
```
EmoGarbage404:
```suggestion
            if (_physics.GetHardCollision(anchored).Layer & (int)placer.Comp.BlockingLayer != 0) // To stop placing wires under walls
```
collision should be using bitmask flag checks

PR #2570, opened 2026-08-22
last marauder updates, hopefully

Flareguy:
## About the PR
updated the marauder cache for the last time hopefully. stimulant pen was not interesting so now they have their grenade back. i also missed the grenade a lot.

## Media

N/A

**Changelog**

:cl:
- tweak: Replaced the marauder's stimulant pen with an explosive grenade.


EmoGarbage404 (changes requested):
what i said in disc

PR #2611, opened 2026-08-25
satchel space now same as backpack

FyeStance:
## About the PR
same storage 

**Changelog**
:cl:
- tweak: Satchels now have the same storage as backpacks


EmoGarbage404:
I don't think there's any point in doing this and having satchels exist alongside backpacks. I'd rather just sunset all instances of satchels and then when storage is a bit more developed we could figure out if it's a meaningful element to bring back for certain situations.

PR #2615, opened 2026-08-26
some mask weight changes

mirrorcult:
## About the PR

way less pickpocket until i fix my rider and can make the changes i wanted
less mercenary since exsyndie
more vandal is fine


EmoGarbage404:
i want to sunset merc at some point but this is fine for now

mirrorcult:
yeah i figured we were doing that at some point anyway so this would be fine

PR #2629, opened 2026-08-27
Superfan Refactor

### LINKED ISSUE #1079: Superfan needs a test asserting they're configured in any given masquerade.
tl;dr if a masquerade can have them, they must be configured, and if it cannot they shouldn't be.

Superfan doesn't need to function in cases where there's no masquerade and doesn't need to function in masquerades that don't support it.

i have a thing in mind i just dont have the energy to code it today.

```
Specifically the test would enumerate the possible set of masks for the masquerade and, if it determines the superfan is reachable, mandate a target.

It should not be reachable in any other way besides forcing it in a masquerade that doesn't have them configured, which I think is valid to consider an error condition but will probably make other tests I want to write (like one for an assailant killing every mask and an equivalent for being a bystander to a killing, and asserting no errors occur) fail when they shouldn't.

tl;dr superfan should not be forced in a masquerade that can't handle it
```

EmoGarbage404:
## About the PR

Closes #1079 in a roundabout way

random cleanup to remove sleeper agent definitions from masquerades. this was kinda a balancing piece that we never really used and it kinda just is silly to be defining over and over.

i didnt test this so uh. dont merge yet. we'll get there.

## Media

N/A

**Changelog**

### DIFF
--- a/Content.Server/_ES/SecretIdentity/Superfan/ESSuperfanComponent.cs
+++ b/Content.Server/_ES/SecretIdentity/Superfan/ESSuperfanComponent.cs
@@ -15,5 +15,5 @@ public sealed partial class ESSuperfanComponent : Component
     public ProtoId<ESOrganizationPrototype> TargetOrganization;
 
     [DataField(required: true)]
-    public MasqueradeEntry TargetSecretIdentity;
+    public MasqueradeEntry? TargetSecretIdentity;
 }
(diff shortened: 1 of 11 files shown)


EmoGarbage404:
working on this reveals some insane initialization order bug that i genuinely can't be fucked to puzzle out for something this small.

PR #2630, opened 2026-08-27
Predicted crew manifest

SpeltIncorrectyl:
## About the PR
Crew manifest is predicted.

New shared CrewManifestComponent which is put on the station and dirtied by the sever-side CrewManifestSystem.
The crew manifest pda app system is moved to shared and reads the component off the station instead of using the CrewManifestSystem.
Also had to fix prediction bugs in CartridgeBoundUserInterface by upgrading to SendPredictedMessage and checking for IsFirstTimePredicted.

:cl:
- tweak: Crew manifest is now predicted

### DIFF
--- a/Content.Client/CartridgeLoader/CartridgeLoaderBoundUserInterface.cs
+++ b/Content.Client/CartridgeLoader/CartridgeLoaderBoundUserInterface.cs
@@ -36,6 +36,9 @@ public abstract class CartridgeLoaderBoundUserInterface : BoundUserInterface
             return;
         }
 
+        if (!_timing.IsFirstTimePredicted)
+            return;
+
         // TODO move this to a component state and ensure the net ids.
         var programs = GetCartridgeComponents(_entManager.GetEntityList(loaderUiState.Programs));
         UpdateAvailablePrograms(programs);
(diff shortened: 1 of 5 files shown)


Content.Client/CartridgeLoader/CartridgeLoaderBoundUserInterface.cs
```diff
@@ -17,10 +18,12 @@ public abstract class CartridgeLoaderBoundUserInterface : BoundUserInterface
     private Control? _activeUiFragment;
 
     private IEntityManager _entManager;
+    private IGameTiming _timing;
```
EmoGarbage404: you can actually just use the dependency attribute

Content.Client/CartridgeLoader/CartridgeLoaderBoundUserInterface.cs
```diff
@@ -17,10 +18,12 @@ public abstract class CartridgeLoaderBoundUserInterface : BoundUserInterface
...
 
     protected CartridgeLoaderBoundUserInterface(EntityUid owner, Enum uiKey) : base(owner, uiKey)
     {
         _entManager = IoCManager.Resolve<IEntityManager>();
```
EmoGarbage404: BoundUserInterface has EntMan on the base class so this isn't necessary (optional sidequest for 2 seconds)

Content.Shared/CartridgeLoader/Cartridges/CrewManifestCartridgeSystem.cs
```diff
@@ -61,7 +61,11 @@ private void UpdateUiState(EntityUid uid, EntityUid loaderUid, CrewManifestCartr
         if (owningStation is null)
             return;
 
-        var (stationName, entries) = _crewManifest.GetCrewManifest(owningStation.Value);
+        if (!TryComp<CrewManifestComponent>(owningStation, out var manifestComp))
+            return;
+
+        var stationName = MetaData(owningStation.Value).EntityName;
+        var entries = manifestComp.Entries;
```
EmoGarbage404, line 68: change GetCrewManifest() to just retrieve this component and read off the data. It wouldn't change the API at all and also would let you get rid of the cached entries on the system.

EmoGarbage404 (changes requested):

PR #2665, opened 2026-09-02
Barnacle sense

### LINKED ISSUE #2643: There should be a way to see the barnacles placements from the character/objective menu
eventually when we get map button it should just be a beacon that only the worm can see, but for now the best approx. would be something like.

You have placed a barnacle near Dorms 4. it is North West of you.
You have placed a barnacle near the Morgue. it is North East of you.

No need to say anything about remaining barnacles probably because the hud exists, but it might be good for the sake of stagehands
tho stagehands wouldnt be able to see it if its done stalker style shrugs

PotatOverlord:
## About the PR

gives the barnacle a record of the locations of their barnacles. it doesnt tell them the direction of said barnacles though which would ideally be done with some kind of alert or pinpointer thingy idk, and this solution isnt very good either because it relies on the user having a mental map + barnacles are usually way off in fuck knows where maints so the beacons don't really matter. The best simplest way would be for there to be a map bind as discussed a few times previously and then have a private beacon to the barnacle of where their guys are

closes #2643 BUT i dont think its a very good solution (obv) but its also better than nothing as everything else has a few pre requisites

## Media

20260902_10h25m00s_grim

**Changelog**

:cl:
- add: Barnacles can now see a record of their barnacley exploits in the character menu.


Resources/Locale/en-US/_ES/secret-identities/parasite/barnacle.ftl
```diff
@@ -1,3 +1,4 @@
 ﻿barnacle-planting-start = Planting barnacle...
 barnacle-max = I can't plant any more!
 barnacle-destroyed = One of my barnacles was destroyed {$location}!
+barnacle-location-character-info-blurb = I have barnacle(s) near {$location}.
```
EmoGarbage404: pass the number of locations as a variable and use a condition to split it between "a barnacle" and "barnacles"

also it would be nice if the names of the locations were highlighted a different color for visibility.

EmoGarbage404 (changes requested):

PR #2672, opened 2026-09-04
Optimize tile fire spread weight picking

vgscrape:
## About
Tile fire neighbor spread was allocating a new `Dictionary<EntityCoordinates, float>` on every `SpreadNeighbors` event.

This keeps a cleared scratch `List<(EntityCoordinates, float)>` on the system and does a weighted pick/remove against that list instead. Scoring and spread behavior are unchanged.

## Test Plan
- [x] Start a tile fire and confirm it still spreads across flammable tiles
- [x] Hot tiles / oxygen gates still affect where it prefers to go
- [x] No obvious perf regression with a large fire

### DIFF
--- a/Content.Server/_ES/TileFires/ESTileFireSystem.cs
+++ b/Content.Server/_ES/TileFires/ESTileFireSystem.cs
@@ -29,9 +29,9 @@ public sealed partial class ESTileFireSystem : ESSharedTileFireSystem
     private static readonly EntProtoId Stage1Fire = "ESTileFire";
 
     /// <summary>
-    /// Scratch buffer for weighted neighbor picks. Cleared each spread so we do not alloc a Dictionary per event.
+    /// Scratch weights buffer. Cleared each spread to avoid allocating a new Dictionary per event.
     /// </summary>
-    private readonly List<(EntityCoordinates Coords, float Weight)> _spreadWeights = new();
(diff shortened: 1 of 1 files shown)


Content.Server/_ES/TileFires/ESTileFireSystem.cs
```diff
@@ -130,18 +135,18 @@ private void OnSpreadNeighbors(Entity<ESTileFireComponent> ent, ref SpreadNeighb
...
         }
 
         while (args.Updates > 0)
         {
-            if (flammable.FireStacks < ent.Comp.MinFirestacksToSpread || weights.Count == 0)
+            if (flammable.FireStacks < ent.Comp.MinFirestacksToSpread || _spreadWeights.Count == 0)
                 return;
 
-            var coords = _random.PickAndTake(weights);
+            var coords = PickAndTakeWeight(_spreadWeights);
```
EmoGarbage404: this function isn't real

vgscrape: Yep - switched it to the real `IRobustRandom.PickAndTake(Dictionary<...>)` helper and just reuse a cleared `_spreadWeights` dict on the system so we still avoid the per-spread alloc.

EmoGarbage404:
Don't send me AI code

PR #2675, opened 2026-09-05
many map2 walls and windows

Flareguy:
## About the PR
adds a lot of stuff for hypersleep barracks + orange walls and window backings because why not

## Media
Content Client_LP2vSkiqzH

**Changelog**
N/A


Sadie-silly:
its hype hype hype !!

Resources/Prototypes/_ES/Entities/Structures/Walls/hypersleep.yml
```diff
@@ -0,0 +1,15 @@
+- type: entity
+  parent: ESWallArmory
```
EmoGarbage404: can we do some more sane parenting for these ? Like at the very least I would prefer if these walls weren't in some strange incestuous parenting system where their indestructible nature was nonobvious

Resources/Prototypes/_ES/Entities/Structures/Windows/hypersleep.yml
```diff
@@ -0,0 +1,16 @@
+# Hypersleep
+- type: entity
+  parent: ESWindowArmory
+  id: ESWindowHypersleep
+  name: hypersleep window
```
EmoGarbage404: should be "sterile hardened window" I would think to match the walls

Resources/Prototypes/_ES/Entities/Structures/Windows/hypersleep.yml
```diff
@@ -0,0 +1,16 @@
+# Hypersleep
+- type: entity
+  parent: ESWindowArmory
```
EmoGarbage404: same not on the parenting: i thought we had better window parenting anyways

Resources/Prototypes/_ES/Entities/Structures/Windows/hypersleep.yml
```diff
@@ -0,0 +1,16 @@
...
+  id: ESWindowHypersleep
+  name: hypersleep window
+  description: Lets you stare into the outside world, outside of your bubble of cryo-safety.
```
EmoGarbage404: we should be consistent about calling it hypersleep imo esp. since cryopods and cryogas and cryohusks are very distinct things

EmoGarbage404 (changes requested):

Flareguy:
added many many many more tiles and stuff from what ive done recently

PR #2688, opened 2026-09-07
Shoves Prototype

EmoGarbage404:
## About the PR

basic bitch impl of shoves. replaces disarms and throws the person back a tile and potentially knocks them if they hit a table. also applies a short amount of slowdown and gun inaccuracy. doesnt properly cleanup dead disarm code

moves wideswing to contextually right click. shoves are now on entity left click.

meant to be a nonlethal disengagement option + potential combo tool.

## Media

N/A

**Changelog**

:cl:
- add: Left-clicking on things when in combat mode now shoves them, knocking them back a small distance, staggering them, and disrupting their aim.
- tweak: Wideswing and normal attack are now both on right click, and depend on context.


Coolsurf6:
I am curious how this works with prediction. Been waiting for an implementation of this for ages since 13, would shoving into walls or really any solid object be able to count rather than just tables?

